PR #1, opened 2021-03-16
DOP-1865: Implement /status endpoint

i80and:
Unlike marian, this only populates `lastSync` if the `verbose=1` query parameter is given. It is also significantly more detailed, showing information like "which properties were updated" and "how many documents were removed"

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -69,6 +69,11 @@ interface DatabaseDocument extends Document {
     includeInGlobalSearch: boolean
 }
 
+interface StatusResponse {
+    manifests: string[]
+    lastSync?: RefreshInfo | null
+}
+
 function generateHash(data: string): Promise<string> {
     const hash = crypto.createHash('sha256')
 
@@ -245,14 +250,18 @@ class Marian {
     index: Index
     client: MongoClient
     db: Db
-    collection: Collection
+    lastRefresh: RefreshInfo | null
+    documents: Collection
+    manifests: Collection
 
     constructor(bucket: string, client: MongoClient) {
         this.index = new Index(bucket)
         this.client = client
+        this.lastRefresh = null
 
         this.db = client.db('search')
-        this.collection = this.db.collection('documents')
+        this.documents = this.db.collection('documents')
+        this.manifests = this.db.collection('manifests')
 
         // Fire-and-forget loading
         this.refresh().then(result => {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -245,14 +250,18 @@ class Marian {
     index: Index
     client: MongoClient
     db: Db
-    collection: Collection
+    lastRefresh: RefreshInfo | null
+    documents: Collection
+    manifests: Collection
```
madelinezec, line 255: does marian store the manifests in a collection of a mongo database currently?

src/index.ts
```diff
@@ -487,8 +504,16 @@ class Marian {
...
+            "manifests": this.index.manifests.map(manifest => manifest.searchProperty),
+        }
+
+        if (parsedUrl.query.verbose) {
+            response.lastSync = this.lastRefresh
+        }
+
         res.writeHead(200, headers)
-        res.end()
+        res.end(JSON.stringify(response))
```
madelinezec, line 516: you love to see it

src/index.ts
```diff
@@ -344,6 +353,11 @@ class Marian {
                 assert.ok(manifest.manifestRevisionId)
 
                 await session.withTransaction(async () => {
+                    await this.manifests.updateOne(
```
madelinezec, line 356: will this PR change the current workflow (deploying, triggers creating new manifest, poking marian so that it refreshes)? or will that still be workflow

madelinezec (approved):
Looks good! just a couple of questions, including: does marian have a testing suite? if it doesn't need one, why is that? how does one check/test their work when on marian?

src/index.ts
madelinezec: if so, how have I never seen it before 😃

i80and: Marian has no interaction with mongodb -- its index is entirely internal.

(also, I actually may remove this collection, since it's only ever written to)

i80and:
> 
> 
> Looks good! just a couple of questions, including: does marian have a testing suite? if it doesn't need one, why is that? how does one check/test their work when on marian?

Marian has a test suite, but this is not marian :D search-transport *should* have tests, and will certainly have tests before going into production, but this is still in the "early evolution from POC" stage.

src/index.ts
i80and: Workflow will be unchanged. We will likely want to replace search-transport with a Realm app at some point in the future, but this is a stopgap to that future goal.

sophstad (approved):
💯

PR #2, opened 2021-04-26
DOP-1115: Add test to verify that database synchronization works

i80and:
This also includes a refactor to make testing easier, mainly moving `SearchIndex` to its own file.

The test contains two states, and tests transitioning the database from state 1 to state 2, and back to state 1 again. Tested mutations:
* Added properties (state 2 contains `charts.json`)
* Added documents (state 2 contains `tutorial/second-tutorial/index.html`)
* Removed properties (state 1->2 removes `bi-connector-v1`)
* Removed documents (state 2->1 removes `tutorial/second-tutorial/index.html`)
* Mutated documents (state 1->2 fixes a typo in `manual.json`: `s/Ap/App`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   ],
   "scripts": {
     "tsc": "tsc",
+    "test": "mocha",
     "search-transport": "node --trace-warnings src/index.js"
   },
   "author": "Andrew Aldridge <andrew.aldridge@mongodb.com>",
@@ -19,10 +20,12 @@
     "mongodb": "3.6.3"
   },
   "devDependencies": {
+    "@types/mocha": "^8.2.2",
     "@types/mongodb": "^3.6.9",
     "@types/node": "^13.13.4",
     "eslint": "^6.8.0",
     "mocha": "^7.1.2",
+    "ts-node": "^9.1.1",
     "typescript": "^4.2.3"
   }
 }
--- a/src/SearchIndex.ts
+++ b/src/SearchIndex.ts
@@ -38,7 +38,7 @@ interface Manifest {
     searchProperty: string
 }
 
-interface DatabaseDocument extends Document {
+export interface DatabaseDocument extends Document {
     manifestRevisionId: string
     searchProperty: string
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
   ],
   "scripts": {
     "tsc": "tsc",
+    "test": "mocha",
```
sophstad, line 11: On some level it may be nice to keep all of our JS tests using Jest, but no problem if you're most familiar with mocha. Not a blocker in any way, just a thought I had :)

src/SearchIndex.ts
```diff
@@ -0,0 +1,328 @@
...
+            }
+        });
+
+        hash.write(data);
+        hash.end();
+    })
+}
+
+async function getManifestsFromS3(bucketName: string, prefix: string): Promise<Manifest[]> {
+    const s3 = new S3({apiVersion: '2006-03-01'})
```
sophstad, line 74: Lol `2006`

src/SearchIndex.ts
```diff
@@ -0,0 +1,328 @@
...
+                manifestRevisionId: hash,
+                searchProperty: searchProperty
+            })
+        }, () => {
+            resolve(manifests)
+        })})
+}
+
+async function getManifests(manifestSource: string): Promise<Manifest[]> {
+    const parsedSource = manifestSource.match(/((?:bucket)|(?:dir)):(.+)/)
```
sophstad, line 150: May be helpful to describe the format we're looking for in `manifestSource` somewhere.

sophstad (approved):
Sweet!

src/SearchIndex.ts
i80and: A fine vintage 😆

package.json
i80and: Hm, setting up jest seems really nontrivial (the getting started guide... doesn't seem to work and throws a SyntaxError on its own config file?), and the intent is for this whole project to just go away in the future, so I'll stick with mocha for now

PR #6, opened 2021-05-20
Set up Docker & Drone

i80and:


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:14-alpine
```
sophstad, line 1: This is good, makes me wonder if the [autobuilder's node installation](https://github.com/mongodb/docs-worker-pool/blob/integration/Dockerfile#L24-L27) could be cleaned up.

sophstad (commented):
The Drone and Docker files make sense to me!

schmalliso (approved):

PR #8, opened 2021-06-28
Don't throw when handling a bad origin in checkAllowedOrigin()

i80and:
Also removes deprecated node API.


mango-db (approved):
LGTM 🚀 

- [x] Updates `checkAllowedOrigin` to handle bad origin + adds test
- [x] Removes legacy `url` API

PR #18, opened 2022-05-19
DOP-3002: Apply minimum searchable criteria on indexing

casthewiz:
### The What
See DOP-3002 for details.

- Adds a basic set of naive criteria to apply on indexable entries that filters out potentially poor quality results.
- Splits out operation composition and document clean-up stanzas into reusable helper functions.
- Adds an `unindexable` collection as a class member of `SearchIndex`.
- Implements filter based on said criteria, and adds relevant logic for upserting and cleaning stale entries that are deemed unsearchable to `unindexable`.

### The Why/The Commiserating
In an ideal world this would also be applied within `mut` to the index as a means of surfacing warnings for pages that generate poor search results. `mut` would then have the responsibility of tagging these results for the indexer to flag, and we wouldn't have to maintain the heuristics here.

In the world we inhabit, however, this is the fastest and easiest means of satisfying DOP-3002 and also providing us with an abstraction we can iterate against until criteria is more... calcified.


casthewiz:
Please ignore the files in the `build/` directory - they're the built artifacts that will be deployed under the current deployment scheme.

package.json
```diff
@@ -2,12 +2,15 @@
...
+  "engines": {
+    "node": "^14"
+  },
   "main": "src/index.js",
   "files": [
     "src/index.js"
   ],
   "scripts": {
-    "build": "rm -r build; tsc",
+    "build": "rm -r build; tsc && npm run format",
```
casthewiz, line 13: It's possible to have lint-rejected files in the build directory - formatting as part of the build steps ensures that none get accidentally commited.

Necessary now, not strictly necessary if we move to build artifacts as part of the CD process.

src/SearchHeuristics/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  const MINIMUM_CONTENT_LENGTH = 40;
+  return doc.text.length >= MINIMUM_CONTENT_LENGTH;
+};
+
+const hasPreview = (doc: Document) => !!doc.preview?.trim();
+
+const isInteractive = (doc: Document) => doc.links?.length >= 2;
+
+const meetsMinimumSearchableStandard = (doc: Document) => {
+  const failsThreshold = !hasPreview(doc) && (!hasMinimumTextThreshold(doc) || !isInteractive(doc));
```
casthewiz, line 22: Preliminary querying of our production index shows this will remove roughly 367 results - without the `isInteractive` check we'd remove 2k results. 
```{preview: '','links.2': {$exists:true},$expr: {$gt: [ { $strLenCP: '$text' }, 40]}}``` 

For now, I'm erring on the side of tunable caution.

File: src/SearchHeuristics/index.ts [hunk 3]
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const hasMinimumTextThreshold = (doc: Document) => {
+  const MINIMUM_CONTENT_LENGTH = 40;
+  return doc.text.length >= MINIMUM_CONTENT_LENGTH;
+};
+
+const hasPreview = (doc: Document) => !!doc.preview?.trim();
+
+const isInteractive = (doc: Document) => doc.links?.length >= 2;
```
rayangler, line 19: Why do we require there to be 2 or more links in a document? Do we want a document to fail the threshold if it has 0 or 1 link but has content?

rayangler (commented):
Mostly just questions from me!

src/SearchIndex.ts
```diff
@@ -369,3 +346,63 @@ export class SearchIndex {
...
+  log.debug(`Removing old documents for ${manifest.searchProperty}`);
+  const deleteResult = await collection.deleteMany(
+    {
+      searchProperty: manifest.searchProperty,
+      manifestRevisionId: { $ne: manifest.manifestRevisionId },
+    },
+    { session }
+  );
+  status.deleted += deleteResult.deletedCount === undefined ? 0 : deleteResult.deletedCount;
+  log.debug(`Removed ${deleteResult.deletedCount} entries from  documents`);
```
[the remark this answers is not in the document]
casthewiz: That would be a good idea! I 'll make some quick edits here.

src/SearchIndex.ts
```diff
@@ -302,58 +306,31 @@ export class SearchIndex {
...
+          // Upsert documents deemed to be searchable, e.g. we want to surface them to users
+          if (searchableUpserts.length > 0) {
+            const bulkWriteStatus = await this.documents.bulkWrite(searchableUpserts, { session, ordered: false });
+            if (bulkWriteStatus.upsertedCount) status.updated.push(manifest.searchProperty);
+          }
+
+          // Upsert documents rejected from being searchable, for diagnostic purposes
+          if (unsearchableUpserts.length > 0) {
+            const bulkWriteStatus = await this.unindexable.bulkWrite(unsearchableUpserts, { session, ordered: false });
+            if (bulkWriteStatus.upsertedCount) status.updated.push(manifest.searchProperty);
```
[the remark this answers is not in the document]
casthewiz: Good catch - this is potentially part of the response object at the [/status](https://docs-search-transport.docs.staging.corp.mongodb.com/status) path; no strict side effects from pushing twice, but certainly a lot of useless noise. 

I'll compose a string with more info here instead of just the searchProperty.

src/SearchHeuristics/index.ts
casthewiz: It's an arbitrary metric I'm using to avoid false negatives in conjunction with text length and preview existence - the majority of our content has at least 1 or 2 links within a page, so on the off chance that a document's text didn't get properly inputted into the document by `mut`, this boolean can be used to prevent false negatives.

It's obviously a really low bar to clear, but I'm intending to err on the side of avoiding false negatives more than letting some drawer results slip through.

rayangler: Ohhh, that makes more sense. Thank you for the explanation!

rayangler (approved):
Looks good!

PR #19, opened 2022-06-06
DOP-3032: adding synonyms to search queries

schmalliso:
**DO NOT MERGE** until we're confident that this won't break everything.

Per the docs, we need to add a 'synonyms' field to the `$search` agg pipeline stage if you wish to use synonyms in your searching. This should do that.

search_synonyms is currently configured in `search-staging.documents`, my intention is to copy that over to prod once we test this, assuming it's possible to test this only in staging. For what it's worth, if you include the synonyms bit in the query and you've not configured synonyms, you get an angry error:

```
MongoServerError: Remote error from mongot :: caused by :: unknown synonym mapping name "search_synonyms"
```

### DIFF
--- a/src/Query.ts
+++ b/src/Query.ts
@@ -244,6 +244,7 @@ export class Query {
       text: {
         query: terms,
         path: 'text',
+        synonyms: 'search_synonyms',
       },
     });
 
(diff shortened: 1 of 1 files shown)


i80and (approved):
this looks good to me. I might suggest that our required index setup be listed in a `README.md` file just so it's documented, since the specific index setup required for this to run is starting to get nontrivial

PR #25, opened 2022-11-14
DOP-3368: restore support for legacy-style manifests

schmalliso:
[Jira Ticket](https://jira.mongodb.org/browse/DOP-3368) Properties that still build with the legacy tooling use mut-0.9.x to generate their search manifests (from the built HTML rather than from the parser-generated AST). Since we would still like non-Snooty-deployed sites to be included in search, this PR restores the `text` field in query construction in docs-search-transport and makes the `code` and `paragraphs` fields optional

### DIFF
--- a/build/Query.js
+++ b/build/Query.js
@@ -220,7 +220,7 @@ class Query {
     parts.push({
       text: {
         query: terms,
-        path: ['paragraphs', 'code.lang'],
+        path: ['paragraphs', 'code.lang', 'code.value', 'text'],
       },
     });
     parts.push({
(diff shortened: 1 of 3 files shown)


build/Query.js
```diff
@@ -220,7 +220,7 @@ class Query {
     parts.push({
       text: {
         query: terms,
-        path: ['paragraphs', 'code.lang'],
+        path: ['paragraphs', 'code.lang', 'code.value', 'text'],
```
schmalliso, line 223: Preliminary testing with code.value being boosted significantly was causing weird code blocks to bubble up aggressively when searching for common parameters like `$and` or `$ne` so am relegating it to the same as content fields.

i80and (approved):
Are the database indexes still inclusive of `test`?

lgtm!

schmalliso:
@i80and The prod index only knows about text (I've not updated it yet for the changes); I'm going to add text back into the staging one after this PR gets merged.

PR #28, opened 2023-01-19
DOP-3441-prod: Upgraded chart versions to 4.12.3, memory limits and requests for prod

mmeigs:
Ticket: [DOP-3441](https://jira.mongodb.org/browse/DOP-3441)

### Notes

Updated memory limits and requests before Kanopy team sets mandatory amounts. Decided to go higher for safety's sake and further investigation of Kanopy's grafana charts over the past six months.

Upgraded `helm-chart` version of prod primarily. However, also upgraded staging to the latest version after speaking with Staff Infra Engineer. This upgrade is necessary for newer releases to follow [k8's 1.22 API scheduled deprecation](https://mongodb.slack.com/archives/CCKCR28BH/p1660591811877929).

Before merging but after approval, I will use the `mapkubeapis` Helm plugin to update deprecated and removed API versions in-place which is required for a successful build with the new `helm-chart` upgrade. The steps and reasonings can be seen [here](https://github.com/10gen/kanopy-docs/blob/master/docs/advisories/2023-01-04_deprecated_APIs.md#post-upgrade-remediation-steps).

Upgrade of staging env went smoothly and as planned yesterday.

### DIFF
--- a/environments/prod.yml
+++ b/environments/prod.yml
@@ -26,6 +26,6 @@ probes:
 
 resources:
   limits:
-    memory: 900Mi
+    memory: 1300Mi
   requests:
-    memory: 1200Mi
+    memory: 1100Mi
--- a/.drone.yml
+++ b/.drone.yml
@@ -48,7 +48,7 @@ steps:
     image: quay.io/mongodb/drone-helm:v3
     settings:
       chart: mongodb/web-app
-      chart_version: 4.12.0
+      chart_version: 4.12.3
       add_repos: [mongodb=https://10gen.github.io/helm-charts]
       namespace: docs
       release: docs-search-transport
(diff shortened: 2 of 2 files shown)


environments/prod.yml
```diff
@@ -26,6 +26,6 @@ probes:
 
 resources:
   limits:
-    memory: 900Mi
+    memory: 1300Mi
   requests:
-    memory: 1200Mi
+    memory: 1100Mi
```
rayangler, line 31: I noticed in the PR for staging, we also bumped down the memory for requests. I'm assuming we had it set a little too high?

rayangler (approved):
LGTM! Just one question for context/my understanding

environments/prod.yml
mmeigs: I am embarrassed to say I originally had them backwards... The limit should be higher than the requests. For prod Kanopy has the "suggested" requests for docs-search-transport as 1.04 GB, and limit to be 1.14 GB. I decided to bump that a bit. [Here is the grafana](https://grafana.corp.mongodb.com/d/a164a7f0339f99e89cea5ab47f9be872/memory-profiler-workloads?orgId=1&var-datasource=default&var-namespace=docs&var-workload=docs-search-transport-web-app&var-type=Deployment&from=now-6M&to=now&refresh=15m) for prod as reference!

PR #31, opened 2023-03-22
DOP-3585: allocate memory for node process. default back to 512 after fix

seungpark:
Increase [node process memory allocation ](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes) to avoid error in ticket.

will follow up with a ticket to identify the area of memory leak and revert this fix

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,4 @@ RUN npm ci --production && chown -R node:node /app
 USER node
 
 EXPOSE 8080
-ENTRYPOINT ["node", "build/index.js"]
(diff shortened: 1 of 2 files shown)


i80and (approved):
I don't know Kanopy's defaults, so we might have to request a kanopy memory increase if this fails to start in staging

seungpark:
> I don't know Kanopy's defaults, so we might have to request a kanopy memory increase if this fails to start in staging

added the increase limit in stage 👍

PR #35, opened 2023-04-05
DOP-3596

seungpark:
- add /build directory to .gitignore. remove existing directory
- add a step to Dockerfile to compile the typescript to JS into /build directory

**note:** A/C states a GH action to build and commit the build directory, but this is a workaround for the kanopy drone to read the Dockerfile and populate the /build directory


src/index.ts
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env ts-node
 'use strict';
 
+console.log('test DOP-3596');
```
casthewiz, line 4:
```suggestion
```

seungpark: woops! remnant from testing GH action to see if it'd commit gitignored files

casthewiz (approved):

PR #38, opened 2023-04-14
DOP-3596: include tsc as part of prod dependencies

seungpark:
from this build log:

https://drone.corp.mongodb.com/mongodb/docs-search-transport/76/1/2

```
> search-transport@0.1.0 build /app
> rm -rf build; tsc && npm run format

sh: tsc: not found
npm ERR! code ELIFECYCLE
npm ERR! syscall spawn
npm ERR! file sh
npm ERR! errno ENOENT
npm ERR! search-transport@0.1.0 build: `rm -rf build; tsc && npm run format`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the search-transport@0.1.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

since we install with npm ci --production

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "basic-logger": "^0.4.4",
     "dive": "^0.5.0",
     "dotenv": "^16.0.3",
-    "mongodb": "^5.1.0"
+    "mongodb": "^5.1.0",
+    "typescript": "^4.2.3"
   },
   "devDependencies": {
     "@types/mocha": "^8.2.2",
@@ -37,7 +38,6 @@
     "lint-staged": "^11.0.0",
     "mocha": "^7.1.2",
     "prettier": "2.3.0",
-    "ts-node": "^9.1.1",
-    "typescript": "^4.2.3"
+    "ts-node": "^9.1.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "src/index.js"
   ],
   "scripts": {
-    "build": "rm -rf build; tsc && npm run format",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -27,7 +27,8 @@
     "basic-logger": "^0.4.4",
     "dive": "^0.5.0",
     "dotenv": "^16.0.3",
-    "mongodb": "^5.1.0"
+    "mongodb": "^5.1.0",
+    "typescript": "^4.2.3"
```
rayangler, line 31: I think `typescript` is generally advocated to be a dev dependency. I don't think it matters too much in the grand scheme of things, but I'm wondering if it would be good practice to instead update the Dockerfile to have a separate build stage where we can install dev dependencies and build.

If this seems like a lot of work and is better done another time, this is fine with me

seungpark: good call @rayangler ! see latest commit

rayangler: Oh, sorry, I was thinking about [Docker's stages](https://docs.docker.com/build/building/multi-stage/#use-multi-stage-builds)! Similar to how the [Autobuilder's](https://github.com/mongodb/docs-worker-pool/blob/master/Dockerfile) has a building stage. I have an in-progress smaller scale [example](https://github.com/10gen/snooty-data-api/blob/b11aa2f6e9d69c8001ca3b47bb27e537ab08ffc1/Dockerfile) for the Snooty Data API.

seungpark: TIL docker 😎

rayangler (approved):
LGTM! 🙏 Hoping this fixes everything

PR #39, opened 2023-04-18
DOP-3578: Move tests to new directories, add to ci

casthewiz:
Moves our test suites into a dedicated unit & integration test directory. 

Adds relevant GH action for running CI, with requisite secrets added to the repository.


tests/integration/search.test.ts
casthewiz, line 1: This file has the `deepStrictEquals` assertions removed - the test suites using them essentially replicate jest snapshot functionality, and they're of questionable value as we move ranking logic around.

seungpark (approved):
this LGTM. tested locally as well as the CI 👨‍🍳

PR #41, opened 2023-04-28
DOP-3567

caesarbell:
This handles the multi-index for `code.value` to return v6.0 branches and other newer branches.


seungpark (approved):
this LGTM

To add context, the search index has been updated on search db `search-staging.documents` to include the 'multi' analyzer option on `code.value` field and has been tested on stage. should also verify this update is on the prod db pre merge

PR #42, opened 2023-05-04
DOP-3569: manage Search Index from docs-search-transport

seungpark:
WIP

**Goals:**
- Manage Search Index from search transport server. This should allow for environments to be in sync as well as programmatically handling index modification with Search Facet Taxonomy

**Testing logs:**
- [first successful log of patching searching index](https://mongodb.splunkcloud.com/en-US/app/search/show_source?sid=1683228951.3923649&offset=121&latest_time=)

**Follow ups:**
- Revoke access for Search Index modifications(?). Updates should be done from transport server only


package.json
```diff
@@ -28,11 +28,14 @@
     "dive": "^0.5.0",
     "dotenv": "^16.0.3",
     "mongodb": "^5.1.0",
+    "node-fetch": "^2.6.11",
```
seungpark, line 31: in hindsight, can remove this since we already make use of of urllib (required for digest Auth fetch with Atlas Admin) to keep it light as possible but this is only few kB https://www.npmjs.com/package/node-fetch/v/2.6.11

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
...
+    } catch (e) {
+      log.error(`Error while updating search index: ${JSON.stringify(e)}`);
+      throw e;
+    }
+  }
+
+  private _getPostPatchOptions(searchIndex: IndexMappings, taxonomy: Taxonomy) {
+    const options = DEFAULT_ATLAS_API_OPTIONS;
+    options['data'] = this._insertTaxonomyIntoSearchIndex(searchIndex, taxonomy);
+    options['data'] = Object.assign(options['data'], {
```
branberry, line 119: Super NIT: Since `COLLECTION_NAME`, `DB`, and `SEARCH_INDEX` are available globally, could the `_insertTaxonomyIntoSearchIndex` add these properties when it returns the object?

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
...
+      ];
+    }
+    return searchIndex;
+  }
+}
+
+// TODO: possibly move to taxonomy utils
+const getFacetKeys = (taxonomy: Taxonomy) => {
+  const keyList: string[] = [];
+  const pushKeys = (baseStr = '', currentRecord: Taxonomy) => {
```
branberry, line 158: Super NIT: If `baseStr` is the second argument for the function, this would allow us to call `pushKeys` on line 171 without the redundant extra string.

src/index.ts
```diff
@@ -295,7 +316,14 @@ async function main() {
     await searchIndex.createRecommendedIndexes();
   }
 
-  const server = new Marian(searchIndex);
+  const atlasAdmin = new AtlasAdminManager(process.env['ATLAS_ADMIN_PUB_KEY']!, process.env['ATLAS_ADMIN_API_KEY']!);
```
branberry, line 319: I think we should check these env vars instead of doing a non-null assertion to ensure that we have these values populated. We could catch the error here and send a helpful message if these env vars are not set.

src/index.ts
```diff
@@ -181,14 +155,33 @@ class Marian {
...
+      throw new InvalidQuery();
+    }
+
+    const query = new Query(rawQuery);
+
+    let searchProperty = parsedUrl.searchParams.getAll('searchProperty') || null;
+    if (typeof searchProperty === 'string') {
+      searchProperty = [searchProperty];
+    }
+    return await this.index.search(query, searchProperty);
```
branberry, line 174: NIT: The `await` is not needed here.

src/index.ts
```diff
@@ -242,6 +235,33 @@ class Marian {
     res.writeHead(200, headers);
     res.end(JSON.stringify(response));
   }
+
+  async load() {
+    let taxonomy: Taxonomy;
+    try {
+      taxonomy = await this.fetchTaxonomy(process.env.TAXONOMY_URL!);
```
branberry, line 242: We should check if the `TAXONOMY_URL` env var is undefined instead of non-null assertion.

branberry (commented):
Hey Seung! This looks good, I just left comments largely regarding TypeScript stuff, and other various small nitpicks. Let me know if you feel like any of my suggestions are more work than warranted for the time being, and I can re-review.

branberry (commented):
Hey Seung! This looks good, I just left comments largely regarding TypeScript stuff, and other various small nitpicks. Let me know if you feel like any of my suggestions are more work than warranted for the time being, and I can re-review.

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  private async findSearchIndex(db: string, collection: string, indexName: string) {
+    log.info('finding Atlas search index');
+    const url = `${this.baseUrl}/${db}/${collection}`;
+    const options = DEFAULT_ATLAS_API_OPTIONS;
+    options['method'] = 'GET';
+
+    try {
+      const { data, res } = await request(url, options);
```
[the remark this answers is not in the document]
seungpark: great suggestion! api docs actually shows same response types for [each request ](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Atlas-Search/operation/createAtlasSearchIndex) so will be adding these to atlas-types.ts and export

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
...
+      }
+
+      return data;
+    } catch (e) {
+      log.error(`Error while updating search index: ${JSON.stringify(e)}`);
+      throw e;
+    }
+  }
+
+  private _getPostPatchOptions(searchIndex: IndexMappings, taxonomy: Taxonomy) {
```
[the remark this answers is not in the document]
seungpark: fair fair. i've denoted the `_` to do some external testing, so will leave this as `private getPostPatchOptions` thanks for catching consistency!

src/AtlasAdmin.ts
seungpark: i'd actually leave this here as this assigns the data for post/patch options. `insertTaxonomyIntoSearchIndex` should literally convert the taxonomy into search index keys

src/index.ts
seungpark: combined with the group_id logic and added a env var verification

src/index.ts
seungpark: removed !

src/index.ts
seungpark: there actually is no taxonomy url at the moment, hence the placeholder data. i will add a comment here to update this into the env variable verification i included above

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
...
+  dataType: 'json',
+};
+
+const log = new Logger({
+  showTimestamp: true,
+});
+
+const CLUSTER_NAME = process.env['CLUSTER_NAME'] || 'Search';
+const COLLECTION_NAME = process.env['COLLECTION_NAME'] || 'documents';
+const GROUP_ID = process.env['GROUP_ID'] || '';
```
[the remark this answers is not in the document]
seungpark: included a `verifyEnvVars` fn to verify all the new env vars :) left out TAXONOMY_URL since it is not available yet, hence the fallback data

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
...
+      if (!index) {
+        return this.createSearchIndex(SearchIndex, taxonomy);
+      }
+      return this.updateSearchindex(index['indexID'], SearchIndex, taxonomy);
+    } catch (e) {
+      log.error(`Error while patching searching index: ${JSON.stringify(e)}`);
+    }
+  }
+
+  private async findSearchIndex(db: string, collection: string, indexName: string) {
```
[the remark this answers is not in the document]
seungpark: done

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
+// @ts-ignore
+import Logger from 'basic-logger';
```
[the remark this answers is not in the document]
seungpark: this is giving me a 404 `npm ERR! 404 Not Found - GET https://registry.npmjs.org/@types%2fbasic-logger - Not found` not sure if there is any typing for this module

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,173 @@
...
+      return data.find((index: any) => index.name === indexName);
+    } catch (e) {
+      log.error(`Error while fetching search index: ${JSON.stringify(e)}`);
+      throw e;
+    }
+  }
+
+  private async createSearchIndex(searchIndex: IndexMappings, taxonomy: Taxonomy) {
+    log.info('creating Atlas search index');
+    const url = `${this.baseUrl}`;
```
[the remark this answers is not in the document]
seungpark: good catch

src/AtlasAdmin.ts
seungpark: Done!

src/AtlasAdmin.ts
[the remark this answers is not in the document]
branberry: ah gotcha, that would make sense lol

src/AtlasAdmin.ts
branberry: Gotcha that makes sense to me!!

src/AtlasAdmin.ts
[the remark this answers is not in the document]
branberry: Gotcha! Since the `verifyEnvVars` function returns the env variables, could we use those to access env vars?  Since it gets the env vars, maybe we could name it `verifyAndGetEnvVars`. That way, TypeScript won't complain as much

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,172 @@
...
+const SEARCH_INDEX = 'default';
+
+/**
+ * Manager is intended to keep Atlas Search Index in sync across environments
+ * Index should be edited in ../data directory and expect to be updated on deploy
+ */
+
+export class AtlasAdminManager {
+  private defaultHeaders: RequestOptions;
+  private baseUrl: string;
```
branberry, line 32: Super NIT: Would it be worth while making `baseUrl` readonly?

src/AtlasAdmin.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  async patchSearchIndex(taxonomy: Taxonomy) {
+    log.info('patchSearchIndex');
+    try {
+      const index = await this.findSearchIndex(DB, COLLECTION_NAME, SEARCH_INDEX);
+      if (!index) {
+        return this.createSearchIndex(SearchIndex, taxonomy);
+      }
+      return this.updateSearchindex(index['indexID']!, SearchIndex, taxonomy);
```
branberry, line 49: Sorry to be a stickler about this, lol, but I think we should check if `indexID` is defined instead of using the non-null assertion operator.

If you do something like

```ts
.
.
.
// index.indexID is of type string | undefined before this point

if (!index.indexID) {
 
 throw new Error('IndexID is not defined');
}

// index.indexID is of type string past this point
.
.
.
```

branberry (commented):
Hey Seung! Looks really good, just a few more small things

src/AtlasAdmin.ts
seungpark: im gonna update the response to always have this instead of making a conditional

https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Atlas-Search/operation/listAtlasSearchIndexes

not sure why the indexId would not be `required` under the response, but the docs show the response will return indexid: https://www.mongodb.com/docs/atlas/atlas-search/tutorial/create-index/

src/AtlasAdmin.ts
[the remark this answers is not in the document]
seungpark: done. passed to AtlasAdmin Manager from index.ts after getting env vars

src/AtlasAdmin.ts
seungpark: done

branberry (approved):
LGTM!

casthewiz (approved):

PR #47, opened 2023-05-30
update tests based on changes to search index server

seungpark:
Two changes broke the unit tests:
- search transport server now instantiates after loading in the manifests (as well as search index update)
- search index requires a `Taxonomy` object when loading (required for query parsing)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       ATLAS_URI: ${{ secrets.CI_ATLAS_URI }}
+      GROUP_ID: ${{ env.GROUP_ID }}
+      ATLAS_ADMIN_API_KEY: ${{ secrets.ATLAS_ADMIN_API_KEY }}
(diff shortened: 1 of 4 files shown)


seungpark:
@casthewiz what worked locally is failing due to environment variables. that said, we will have to introduce something like sinon to mock `update atlas search index` to **not** update search indexes in the tests while loading the index because that is the added logic of `index.load`

```
  http interface
Missing GROUP_ID
Missing ATLAS_ADMIN_API_KEY
Missing ATLAS_ADMIN_PUB_KEY
Usage: search-transport [--create-indexes]

The following environment variables are used:
* MANIFEST_URI
* ATLAS_URI
* ATLAS_DATABASE (defaults to "search")
* GROUP_ID
* ATLAS_ADMIN_API_KEY
* ATLAS_ADMIN_PUB_KEY 
```

casthewiz:
@seungpark We can use github env variables or secrets to pipe in relevant values into the action. 

Agreed that the unit tests shouldn't be updating the actual db instance if possible. 
The integration tests *should* update the `search-test` db and query against it as much as possible imo - so I wouldn't go to using `sinon` or other mocks there.

seungpark:
> 

that actually sounds like a great idea. will add these into env secrets

PR #50, opened 2023-06-12
Minor update to readme

seungpark:
Added a minor bit of components used in search server

[Trivial follow up task](https://jira.mongodb.org/browse/DOP-3771) to separate the Marian server code.

Pls comment if anything else can be cleared up!

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ REST server that indexes search documents within an Atlas cluster and handles us
 ## Components
 
 ### Marian Server
-REST server that exposes endpoints to run queries against A DB.
+REST server that exposes endpoints to run queries against Atlas DB.
 
 ### Atlas Admin Manager
 Module to invoke Atlas Admin commands via the [Atlas Admin API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/), specifically manages the Atlas Search Index for Search DB.
--- a/README.md
+++ b/README.md
@@ -2,6 +2,17 @@
 
 REST server that indexes search documents within an Atlas cluster and handles user search requests, deployed via Kanopy
 
+## Components
+
+### Marian Server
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,17 @@
 
 REST server that indexes search documents within an Atlas cluster and handles user search requests, deployed via Kanopy
 
+## Components
+
+### Marian Server
+REST server that exposes endpoints to run queries against A DB.
+
+### Atlas Admin Manager
+Module to invoke Atlas Admin commands via the Atlas Admin API.
```
rayangler, line 11: (Non-blocking suggestion) Would it be a good idea to include a sentence to provide an overview/more context on what the purpose of these commands are for? Something like `"The module syncs the Search Index to ensure content taxonomy is incorporated in faceted search data."` ?

README.md
```diff
@@ -2,6 +2,17 @@
 
 REST server that indexes search documents within an Atlas cluster and handles user search requests, deployed via Kanopy
 
+## Components
+
+### Marian Server
+REST server that exposes endpoints to run queries against A DB.
```
rayangler, line 8: (Super duper nit) Should we lowercase "A DB" to "a DB"? Or does the "A" stand for "Atlas" in this case?

rayangler (commented):
Looks good! Just wanted to learn more about the Atlas Admin Manager and was wondering if an extra sentence might help with context. If unnecessary, I'm okay with dropping it!

README.md
seungpark: Atlas DB is correct :) thanks for the catch!

README.md
seungpark: good suggestion! i didn't want to explain specific details of projects so i added the admin API, as well as the Search Index. lmk if this works!

rayangler (approved):

PR #51, opened 2023-06-21
[DOP-3711]: Create/update synonyms in Docs Search Transport

branberry:
## Ticket:
[DOP-3711](https://jira.mongodb.org/browse/DOP-3711)


src/AtlasAdmin.ts
```diff
@@ -1,9 +1,12 @@
 // @ts-ignore
 import Logger from 'basic-logger';
 import { request, RequestOptions } from 'urllib';
+import fs from 'fs';
+import path from 'path';
 import { SearchIndex } from './data/atlas-search-index';
-import { SearchIndexResponse, IndexMappings } from './data/atlas-types';
+import { SearchIndexResponse, IndexMappings, SynonymDocument } from './data/atlas-types';
 import { Taxonomy } from './SearchIndex';
+import { AnyBulkWriteOperation, MongoClient } from 'mongodb';
```
rayangler, line 9: (Nit) Can we import above relative imports for organizational purposes?

src/AtlasAdmin.ts
```diff
@@ -49,6 +54,25 @@ export class AtlasAdminManager {
...
+
+    const synonymUpdates = getSynonymUpdateOperations('../synonyms.csv');
+
+    console.log('uploading the following parsed synonyms documents to Atlas: ', synonymUpdates);
+    try {
+      // we want to ensure that the primary property is a unique index
+      // so that we prevent duplicate synonym records.
+      // the createIndex method is idempotent, so calling it repeatedly will not cause side effects.
+      await synonymCollection.createIndex({ primary: 1 }, { unique: true });
+      await synonymCollection.bulkWrite(synonymUpdates);
```
rayangler, line 70: Does order of the updates matter? If not, I'd recommend making this an unordered bulk write since it can be slightly faster (in theory), especially if we have a lot of updates to perform

src/index.ts
```diff
@@ -229,10 +229,15 @@ class Marian {
     try {
       // TODO: include taxonomy url in verifyEnvVars after it has been released
       taxonomy = await this.fetchTaxonomy(process.env.TAXONOMY_URL!);
-      const atlasAdminRes = await this.atlasAdmin.patchSearchIndex(taxonomy);
-      const loadRes = await this.index.load(taxonomy);
+      await this.atlasAdmin.updateSynonyms();
+
+      const [atlasAdminRes, loadRes] = await Promise.all([
```
rayangler, line 234: Just wondering if responses from these two need to be saved? Looks like they're unused

rayangler (commented):
Not an expert on docs search transport, but this looks pretty good! Just a few comments from me

src/index.ts
branberry: I don't believe they are used now, I just modified the [existing code](https://github.com/mongodb/docs-search-transport/blob/dc0de2e5255cf959ffbcb22551b0aab6fea4c26c/src/index.ts#L233) to do these requests in parallel while keeping the results stored as a variable. Not sure if they are planned on being used in the future, but I kept them in just in case. Would you prefer if I remove them now and just add them in later as needed?

rayangler: My general preference is to not have them if we don't use them/plan on using them in order to avoid any potential confusion. But @seungpark would probably have more context on this!

rayangler (approved):
"LGTM" would be a synonym for this PR 😎 (I know this doesn't make any sense)

src/AtlasAdmin.ts
```diff
@@ -49,6 +54,25 @@ export class AtlasAdminManager {
...
   }
 
+  async updateSynonyms(): Promise<void> {
+    console.log('Updating synonyms');
+
+    const synonymCollection = this.mongoClient.db(DB).collection<SynonymDocument>(SYNONYM_COLLECTION_NAME);
+
+    const synonymUpdates = getSynonymUpdateOperations('../synonyms.csv');
+
+    console.log('uploading the following parsed synonyms documents to Atlas: ', synonymUpdates);
```
seungpark, line 64: can remove this log

seungpark (approved):
this LGTM 👍 i think we can expand on the `analyzer` [option](https://github.com/mongodb/docs-search-transport/pull/51/files#diff-d6230d22597a0dbadc726e7b7b081aced02ab9c7527af0634fe50e8b78a0af68R90) after we further test search results

PR #53, opened 2023-07-07
[DOP-3838]: Update dockerignore to allow synonyms.csv

branberry:


seungpark (approved):
thanks @branberry for the quick response!

PR #54, opened 2023-07-13
[DOP-3838]: Move synonyms to root directory

branberry:


seungpark (approved):
this LGTM. minor offline debate for /resources vs /src/data but no huge preference 👍 as long as its copied over in Dockerfile

PR #55, opened 2023-07-14
DOP-3835: return formatted response for v2/status

seungpark:
**Goal:** to provide a Trie structure of available facets to front end. This will be in the same format as the returned response from /v2/search/meta so that front end can easily determine hierarchy as well as reference display names.

sample response can be found [here](https://docs-search-transport.docs.staging.corp.mongodb.com/v2/status)

The trie structure is a master hash trie, with some special attributes that front end will have to be aware of (ie. `versions`, `name`, `stable` flag for versions etc.) [At first](https://github.com/mongodb/docs-search-transport/pull/55/commits/005278d1a436bc54c2021101698aff8eca496297) it was a list of facets with nested lists, but this causes a array lookup in each layer (front end, search index), which can lead to performance issues.

This hash trie will be used in DOP-3859 to build basic filters on requests to /v2/search/meta (ie. expand all hierarchical facets)


src/util.ts
```diff
@@ -15,3 +17,51 @@ export function arrayEquals<T>(arr1: Array<T>, arr2: Array<T>): boolean {
...
+ * each node is denoted by a 'name' attribute.
+ * other attributes denotes a new node
+ * ['name' and 'display_name' attributes denote name(s) of facet]
+ * [versions have special boolean attribute of 'stable']
+ */
+export function convertTaxonomyResponse(taxonomy: Taxonomy): FacetDisplayNames {
+  const res: FacetDisplayNames = {};
+
+  function addToRes(entityList: TaxonomyEntity[], ref: { [key: string]: any }, property: string) {
+    const conversion = property === 'versions' ? (s: string) => s : convertTitleCase;
```
casthewiz, line 38: I would strongly consider moving this logic into `convertTitleCase` so we don't have a dynamic function definition - can be hard to trace and isn't very obvious on first glance

src/util.ts
```diff
@@ -15,3 +17,51 @@ export function arrayEquals<T>(arr1: Array<T>, arr2: Array<T>): boolean {
...
+        name: taxEntity['display_name'] || conversion(taxEntity['name']),
+      };
+      if (property === 'versions' && taxEntity['stable']) {
+        entity['stable'] = true;
+      }
+      for (const key in taxEntity) {
+        if (!Array.isArray(taxEntity[key])) {
+          continue;
+        }
+        addToRes(taxEntity[key] as TaxonomyEntity[], entity, key);
```
casthewiz, line 54: I don't think we need the continue here - should be able to do something like this

```suggestion
        if (Array.isArray(taxEntity[key])) {
          addToRes(taxEntity[key] as TaxonomyEntity[], entity, key);
        }
```

seungpark: this was just preference of `if true, return early` (in this case `continue` since its a loop) (and a minor point of saving an indentation 😛 )

```
const fn = () => {
   if (true) {return;}
   doSomething();
   doAnother();
}
```
vs
```
const fn = () => {
   if (false) {
      doSomething()
      doAnother();
   }
}
```

src/util.ts
seungpark: good catch! looking back, reads like an english sentence 😂

seungpark: done

tests/resources/utils-data.ts
```diff
@@ -0,0 +1,94 @@
...
+  ],
+} as Taxonomy;
+
+export const sampleFacetTrie = {
+  genres: {
+    name: 'Genres',
+    reference: { name: 'Reference' },
+    tutorial: { name: 'Tutorial' },
+  },
+  target_platforms: {
```
rayangler, line 39: Just curious if we're intentionally leaving `target_platforms` as snake case, or does it not matter for this field?

tests/unit/util.test.ts
```diff
@@ -1,5 +1,7 @@
-import { strictEqual } from 'assert';
-import { isPermittedOrigin, arrayEquals } from '../../src/util';
+import { strictEqual, deepStrictEqual } from 'assert';
+import { isPermittedOrigin, arrayEquals, convertTaxonomyResponse } from '../../src/util';
+import { Taxonomy } from '../../src/SearchIndex';
```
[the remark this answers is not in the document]
seungpark: removed in the follow up PR!

tests/resources/utils-data.ts
seungpark: this was intentionally left as snake case to coincide with the original taxonomy.toml. snake case should be consistent across the chain of process along taxonomy -> search index -> aggregation queries -> query parameters

src/index.ts
```diff
@@ -306,6 +310,19 @@ class Marian {
...
+
+  private async handleStatusV2(req: http.IncomingMessage, res: http.ServerResponse) {
+    const headers: Record<string, string> = {
+      'Content-Type': 'application/json',
+      Vary: 'Accept-Encoding, Origin',
+      'Cache-Control': 'public,max-age=120,must-revalidate',
+    };
+    Object.assign(headers, STANDARD_HEADERS);
+    checkAllowedOrigin(req.headers.origin, headers);
+    let responseBody = JSON.stringify(this.index.convertedTaxonomy);
```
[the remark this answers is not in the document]
seungpark: good catch. updated

src/index.ts
```diff
@@ -404,4 +421,5 @@ try {
   main();
 } catch (err) {
   console.error(err);
+  process.exit(1);
```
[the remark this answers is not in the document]
seungpark: good catch! try catch will only work in async functions. updated

rayangler (approved):

PR #56, opened 2023-07-25
DOP-3859 separate endpoints

seungpark:


src/Query.ts
```diff
@@ -294,222 +297,203 @@ export class Query {
...
-    });
-    parts.push({
-      text: {
-        query: terms,
-        path: 'tags',
-        score: { boost: { value: 10 } },
-      },
-    });
+  getAggregationQuery(searchProperty: string[] | null): any[] {
+    // getAggregationQuery(searchProperty: string[] | null, facetFilters: FacetFilters): any[] {
```
rayangler, line 349: Should we remove this commented code?

src/Query.ts
```diff
@@ -294,222 +297,203 @@ export class Query {
...
+  const filter: Filter<Document> = {};
+  for (const [key, value] of searchParams) {
+    if (!key.startsWith('facets.')) {
+      continue;
     }
+    const facetNames = key.replace('facets.', '').split('>');
+    for (let facetIdx = 0; facetIdx < facetNames.length; facetIdx++) {
+      // hierarchy facets denoted by >
+      // each facet node requires a facet property, at every even level
+      if (facetIdx % 2 === 1) continue;
```
rayangler, line 390: (Nit) I liked your logic in the other `for` loop where we just increment by 2. Is that something we can do here as well? (but keep the comment, since it's good context)

src/Query.ts
```diff
@@ -294,222 +297,203 @@ export class Query {
...
+      const facetName = facetNames[facetIdx];
+      // construct partial facet name
+      const prefix = facetIdx === 0 ? '' : facetNames.slice(0, facetIdx).join('>') + '>';
+      const facetKey = `facets.${prefix}${facetName}`;
+      // const facetKey = `facets.${facetName}`;
```
rayangler, line 395: Commented code?

src/index.ts
```diff
@@ -260,12 +262,13 @@ class Marian {
 
       throw err;
     }
-    let responseBody = JSON.stringify(results[0]);
+
+    let responseBody = JSON.stringify(results);
```
rayangler, line 266: (Super nit) `const` instead of `let` here?

src/index.ts
```diff
@@ -277,19 +280,14 @@ class Marian {
       throw new InvalidQuery();
     }
 
-    const query = new Query(rawQuery);
+    let filters = extractFacetFilters(parsedUrl.searchParams);
```
rayangler, line 283: (Super nit) `const` instead of `let` here?

rayangler (commented):
Mostly nits/cleanup, and a couple of questions

src/Query.ts
```diff
@@ -294,222 +297,203 @@ export class Query {
...
-        : { includeInGlobalSearch: true };
-
-    const compound: { should: any[]; must?: any[]; minimumShouldMatch: number } = {
+    const compound: { should: any[]; must?: any[]; filter?: any[]; minimumShouldMatch: number } = {
       should: parts,
       minimumShouldMatch: 1,
     };
 
+    if (Object.keys(this.filters).length) {
+      compound['filter'] = compound['filter'] || [];
```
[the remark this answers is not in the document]
seungpark: at the time being, no. this is the first of the case, but wanted to keep it flexible for any future additional `compound.filter` options

src/Query.ts
seungpark: removed!

src/Query.ts
seungpark: removed

src/index.ts
seungpark: done!

rayangler (approved):
LGTM!

src/Query.ts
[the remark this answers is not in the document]
rayangler: Cool, thanks for the context! Figured it was worth asking about 😄

src/Query.ts
```diff
@@ -294,222 +297,200 @@ export class Query {
...
-    const compound: { should: any[]; must?: any[]; minimumShouldMatch: number } = {
+    const compound: { should: any[]; must?: any[]; filter?: any[]; minimumShouldMatch: number } = {
       should: parts,
       minimumShouldMatch: 1,
     };
 
+    // if (Object.keys(this.filters).length) {
+    //   compound['filter'] = compound['filter'] || [];
+    //   constructAggFilters(compound['filter'], this.filters);
+    // }
```
rayangler, line 308: (Just curious) If we don't plan on supporting this, should we remove the code?

rayangler (commented):
@seungpark Sounds good, still lgtm. Leaving one comment to ask about the commented out code, mostly because I'm not sure what the plan is for it

src/Query.ts
seungpark: good point. i dont want filters. i'll remove the remnant filter logic and add you to PR 👍

PR #63, opened 2023-08-10
DOP-3787: Adding pagination to search

bianca-laube:
## Stories/Links:
[DOP-3787](https://jira.mongodb.org/browse/DOP-3787)

## Description:
This PR allows the existing route (https://docs-search-transport.mongodb.com/search) to take in a new query parameter `page` and paginate the results as necessary

Changes:
- Introduce a pool db connection and a new searchPropertyMapping module that [mimics what the app services function was doing](https://realm.mongodb.com/groups/5bad1d3d96e82129f16c5df3/apps/5d41ef5e2de84772fc7c2de2/functions/6279548c1fdab5262a3460c8). This allows the search server to have the search property mapping in memory in order to pass these into the aggregation query (see query link below)
- Allow a page query parameter and paginate the results accordingly
- Requests with no page query parameter will still be limited to 50 results (as existing), so front end does not break with this change.

**NOTES:**
-  For the existing route (without page query parameter) the filtering by `search.documents[].searchProperty` is being done in the query as shown [here (see `$match` portion of aggregation query) ](https://mongodb.splunkcloud.com/en-US/app/search/show_source?sid=1692218916.4995519&offset=0&latest_time=). This is intended so that we can get rid of the front end filtering. **In theory, this makes the front-end filtering a no-op**
- some `search.documents[]` are inconsistent in stage and prod, and hence will not show in on stage. please point out these cases so we can manually verify these are due to stage/prod differences
- the app service call to `fetchSearchPropertyMapping` was returning this extra object, removed from the new searchProperty module
```
projectToSearchMap : {cloud: "mms-cloud", onprem: "mms-onprem", cloudgov: "AtlasGov", docs: "manual", cloud-docs: "atlas",…}
```

## Staged changes and logs:
[sample stage search results](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=onprem)
[sample stage search results with pagination](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=onprem&page=1)
[search property mapping logged on stage](https://mongodb.splunkcloud.com/en-US/app/search/show_source?sid=1692219356.4996639&offset=0&latest_time=)


src/SearchPropertyMapping/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    oldGenDocs.forEach((repo) => {
+      parseRepoForSearchProperties(searchPropertyMapping, repo);
+    });
+
+    const repos = await collection.find(query).toArray();
+
+    repos.forEach((r) => {
+      const repo = {
+        project: r.project,
+        search: !r.search
```
rayangler, line 131: Wouldn't `search` always exist given the query `{ search: { $exists: true } }`?

src/SearchPropertyMapping/index.ts
```diff
@@ -0,0 +1,156 @@
...
+        name: 'v5.0',
+        active: true,
+        urlSlug: 'current',
+        versionSelectorLabel: 'Version 5.0 (current)',
+      },
+      {
+        name: 'v4.4',
+        active: true,
+        versionSelectorLabel: 'Version 4.4',
+      },
```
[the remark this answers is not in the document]
seungpark: i am pro inserting into db but i just want to verify that repos_branches shouldn't have any non-xxxx entries. but guessing from this [jira](https://jira.mongodb.org/browse/DOP-3894) we can use `internalOnly` for this. wdyt

src/SearchPropertyMapping/index.ts
seungpark: good catch! removed this ternary

src/SearchPropertyMapping/index.ts
```diff
@@ -0,0 +1,156 @@
...
+      categoryTitle,
+      versionSelectorLabel: versionLabel,
+    };
+  });
+};
+
+// Look at document of a repo to create search properties for the mapping
+const parseRepoForSearchProperties = (searchPropertyMapping: SearchPropertyMapping, repo: Repo) => {
+  let categoryName = repo.search?.categoryName ?? repo.project;
+  let categoryTitle = repo.search?.categoryTitle ?? '';
```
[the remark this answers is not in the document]
seungpark: updated!

src/SearchPropertyMapping/index.ts
```diff
@@ -0,0 +1,156 @@
...
+          : {
+              categoryTitle: r.search.categoryTitle,
+              categoryName: r.search.categoryName ? r.search.categoryName : null,
+            },
+        branches: r.branches,
+      };
+      parseRepoForSearchProperties(searchPropertyMapping, repo);
+    });
+
+    console.log('the search property is ', searchPropertyMapping);
```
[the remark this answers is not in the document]
seungpark: X'ed the log

src/SearchPropertyMapping/index.ts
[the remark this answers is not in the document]
rayangler: I think that makes sense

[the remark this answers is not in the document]
seungpark: manually inserted into pool_test for now to make sure nothing breaks:

image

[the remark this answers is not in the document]
seungpark: also removed the hardcoded portion and able to return the onprem docs: https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=mms&page=1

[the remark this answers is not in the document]
rayangler: Thanks! I took the liberty of updating the `name` field to `gitBranchName` just to enforce not needing to rely on `name` anymore 👍

src/SearchPropertyMapping/index.ts
```diff
@@ -0,0 +1,102 @@
...
+    // Populate mapping with oldgen docs repos that we might not currently have documents for in the repos_branches collection.
+
+    const repos = await collection.find(query).toArray();
+
+    repos.forEach((r) => {
+      const repo = {
+        project: r.project,
+        search: r.search,
+        branches: r.branches,
+      };
```
rayangler, line 86: It might be worth having this as a projection instead of needing to make a subset of the object.

Maybe something like:

```js
const repos = await collection.find(query).project<Repo>({ _id: 0, project: 1, search: 1, branches: 1 }).toArray();
```

src/SearchPropertyMapping/index.ts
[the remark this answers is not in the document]
seungpark: thanks @rayangler !!

src/SearchPropertyMapping/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  const client = await MongoClient.connect(poolAtlasUri);
+  const db = client.db(dbName);
+  const collection = db.collection<Repo>(collectionName);
+  const query = {
+    search: { $exists: true },
+  };
+  const searchPropertyMapping = {};
+
+  try {
+    // Populate mapping with oldgen docs repos that we might not currently have documents for in the repos_branches collection.
```
rayangler, line 77: (Nit) We can probably remove this comment now that we have some semblance of oldgen docs in the db. Just need to be mindful of adding it to the prod db if they're not there already

rayangler (approved):
LGTM!

src/SearchIndex/index.ts
```diff
@@ -38,9 +38,10 @@ export class SearchIndex {
     this.convertedTaxonomy = {};
   }
 
-  async search(query: Query, searchProperty: string[] | null) {
+  async search(query: Query, searchProperty: string[] | null, pageNumber?: number) {
     const aggregationQuery = query.getAggregationQuery(searchProperty);
-    aggregationQuery.push({ $limit: 50 });
+    const RES_COUNT = 50;
+    const PAGINATED_RES_COUNT = 10;
     aggregationQuery.push({
```
mmeigs, line 45: I think we decided on 250? Correct me if I'm wrong. And then maybe 50 a page... I believe.

mmeigs (approved):
Yay! LGTM. Just a comment about the numbers.

src/SearchIndex/index.ts
seungpark: i will verify with design for results per page! `50 max limit` is whats currently existing (wont change), but a paginated requests should be 10 per page according to design

src/SearchPropertyMapping/index.ts
seungpark: will verify it is in prod db before merge

src/SearchIndex/index.ts
mmeigs: Ah ok, I thought Cassidy had mentioned upping that now that we had pagination. But this can be an easy change in the future if we want!

seungpark: discussed offline with Design! think we overlooked the `items per page` option in the [LG/pagination component](https://www.mongodb.design/component/pagination/example/), so we can do a followup to allow items per page as well as page in the future

https://jira.mongodb.org/browse/DOP-3787?focusedCommentId=5648099&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-5648099

mmeigs: Great, thanks!!

src/SearchPropertyMapping/index.ts
seungpark: image

seungpark: ^ inserted into prod db pool.repos_branches before merge!

PR #65, opened 2023-08-17
DOP-3950: Add emptychar stopword to $ analyzer

casthewiz:
### Context

https://jira.mongodb.org/browse/DOP-3950

This change adds `''` as a stopword to prevent the custom whitespace analyzer being used for `$` prefixed cases from querying for empty strings.


seungpark (approved):
🤞

PR #66, opened 2023-08-24
Update synonym analyzer

sarahemlin:
Hello!

Charles Burnett, our Atlas Search SA, told me our synonym file isn't working because the synonym analyzer needs to match index/query analyzers.  Previously our synonym analyzer was lucene.keyword, but we're using lucene.english for our query & index analyzers.  They both need to be lucene.english for the synonym file to take effect.  I've made that change with this pull request, which should solve the problem.

Thank you!
Sarah

### DIFF
--- a/src/data/atlas-search-index.ts
+++ b/src/data/atlas-search-index.ts
@@ -87,7 +87,7 @@ export const SearchIndex: IndexMappings = {
       source: {
         collection: SYNONYM_COLLECTION_NAME,
       },
-      analyzer: 'lucene.keyword',
+      analyzer: 'lucene.english',
     },
   ],
   analyzers: [


seungpark (approved):
good catch. most fields are indexed with `lucene.english` analyzers, but it seems [code.fields.value](https://github.com/mongodb/docs-search-transport/blob/main/src/data/atlas-search-index.ts#L20-L26) and [code.fields.lang](https://github.com/mongodb/docs-search-transport/blob/main/src/data/atlas-search-index.ts#L11) are whitespace and standard respectively. do we need to set more synonyms with these analyzers with the same source collection?

sarahemlin:
@seungpark I don't think we need to worry about the code matches because we're really concerned about the text and the queries for synonyms.

casthewiz (approved):

seungpark:
reverting this merge commit as its causing a problem with error:
`The following synonym source collections contain an invalid document: synonyms`

My theory is that since we are using lucene.english, synonyms containing punctucation `$` and `$$` are both getting tokenized into empty white space, and this is invalid. Tested on search-test and threw this error when only one document with those values were added.

As for the original problem, we should be including `synonyms` as part of the $search aggregation for each field it is applicable for: https://www.mongodb.com/docs/atlas/atlas-search/text/#synonyms-examples

PR #68, opened 2023-08-30
DOP-3978

seungpark:
**Context:**

Currently, we are getting mismatches on [meta searches](https://docs-search-transport.mongodb.com/v2/search/meta?q=potato) and [search results](https://docs-search-transport.mongodb.com/search?q=potato). The count should match up exactly with the total count of the search results.

**Problem:**
The issue lied with that we were not filtering the $searchMeta query with 
1) search manifest fields defined in DOP-3787 which filters on `document.searchProperty`
2) were not filtering for `includeInGlobalSearch = true`

**Solution:**
Since
1) $searchMeta can only be run as the first step in an aggregation and matching is not possible thereafter, and 
2) $match is [slow performant](https://www.mongodb.com/docs/atlas/atlas-search/performance/query-performance/#-match-aggregation-stage-usage) after a $search stage

Indexed fields `searchProperty` (string) and `includeInGlobalSearch` (boolean) and included them in `compound` operator of both `$search` and `$searchMeta`

[Stage search results shows 1](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=potato)
[Stage search meta results 1](https://docs-search-transport.docs.staging.corp.mongodb.com/v2/search/meta?q=potato)

[Stage search results show 668 (67 pages * 10 + 8 res on last page)](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=test&page=67)
[Stage search meta results shows 668 count](https://docs-search-transport.docs.staging.corp.mongodb.com/v2/search/meta?q=test)

[Stage search works with searchProperty included](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=test&searchProperty=manual-v6.0)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,6 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
-      - 'refs/heads/DOP-3978'
 
 steps:
   - name: publish-staging
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
-      - 'refs/heads/DOP-3787'
+      - 'refs/heads/DOP-3978'
 
 steps:
   - name: publish-staging
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -105,11 +105,12 @@ export default class Marian {
     } catch (err) {
       if (err instanceof InvalidQuery) {
         res.writeHead(400, headers);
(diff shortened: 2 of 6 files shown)


.drone.yml
```diff
@@ -8,7 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
-      - 'refs/heads/DOP-3787'
+      - 'refs/heads/DOP-3978'
```
seungpark, line 11:
```suggestion
```

reminder to self: remove before merge :)

PR #71, opened 2023-09-08
DOP-3962: Add term <-> slug mapping and boost logic

casthewiz:
### Ticket
DOP-3962

### Context
Some functionality, some housekeeping (I really just wanted to use `replaceAll()`).
This PR does multiple things:

#### Functionality
1. Adds a new field indexed to `lucene.keyword`, `strippedSlug`, which contains the value of `slug`, stripped of `\` characters.
2. Adds a query term <-> slug map, allowing for non-DOP contributors to curate matches for specific terms via a page's slug (this is compatible with version filtering)
3. Adds a conditional keyword match query on `strippedSlug` if a `rawQuery` matches a mapped result entry, with score boost by `100` to ensure that mapped results are at the top of total results.

#### Housekeeping
4. Upgrades `docs-search-transport` to Node 18, and switches us to use `ESNEXT` instead of `ES2020`.
5. Upgrades `husky` and adds a short-circuit to prevent `husky` from installing when launching the transport server in KANOPY. This is similar to the [shortcircuit](https://github.com/mongodb/snooty/blob/2d81830a3436e23391a791bca4fa69381eacbb63/package.json#L47) present in `snooty`.

#### Staging Links

Various queries:

##### Basic query for `in`
https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=in

##### Version + Product filter for `in`
https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=in&page=1&searchProperty=manual-v5.0

##### Query for whitespace'd term (`in operator`) that matches a curated result
https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=in%20operator&page=1&searchProperty=manual-v5.0


casthewiz:
`package-lock.json` is larger by virtue of going to `"lockfileVersion": 2`.

.drone.yml
```diff
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOP-3962'
```
casthewiz, line 11:
```suggestion
```

src/data/atlas-search-index.ts
```diff
@@ -56,6 +56,12 @@ export const SearchIndex: IndexMappings = {
         searchAnalyzer: 'lucene.english',
         type: 'string',
       },
+      strippedSlug: {
+        analyzer: 'lucene.keyword',
+        searchAnalyzer: 'lucene.keyword',
+        store: true,
```
seungpark, line 62: what is this `store` option?  [atlas search docs ](https://www.mongodb.com/docs/atlas/atlas-search/stored-source-definition/)reads `doesn't index stored fields and so you can't query these fields, but you can retrieve stored fields at query time using the [returnStoredSource](https://www.mongodb.com/docs/atlas/atlas-search/return-stored-source/#std-label-fts-return-stored-source-option) option`

seungpark (commented):
minor comment about `store` option on index and running string replace on all queries! otherwise LGTM 👍 the search results look dandy 🙌

package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "search-transport",
   "version": "0.1.0",
-  "description": "a tiny search engine",
+  "description": "a not so tiny search engine",
```
mmeigs, line 4: <3

mmeigs (approved):
Interested in Seung's questions, but LGTM

rayangler (approved):

src/data/atlas-search-index.ts
casthewiz: I don't actually remember why I put this here, so I'm removing and testing without the option 🤔

src/Query/index.ts
```diff
@@ -65,6 +66,17 @@ export class Query {
     const parts: any[] = [];
     const searchPropertyMapping = getPropertyMapping();
 
+    // if we need to boost for matching slug on an exact rawQuery match
+    if (resultMapping[this.rawQuery.trim()]) {
+      parts.push({
+        text: {
+          path: 'strippedSlug',
+          query: resultMapping[this.rawQuery.trim()].map((r) => r.replaceAll('/', '')),
```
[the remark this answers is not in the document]
casthewiz: Good catch - I've replaced with a one liner that exports a stripped version of the mapping 👍

seungpark (approved):
this LGTM 🚀

PR #72, opened 2023-09-13
DOCSP-32881: Add more mapped results

casthewiz:
### Context
https://jira.mongodb.org/browse/DOCSP-32881

Adds more mapped results 🥇 💯 

### Pitfalls

`mongocli` has a lot of generic slugs mapped to it - and the first entry in the list doesn't match the entry DOCSP-32881, because its slug would be `''` and would match to every landing page for every product.

Theoretically this might actually be fine, since the rest of the boosts should place the `mongocli` docset pages to the top? 
For the sake of expediency, I've replaced with `command/mongocli` to avoid any collision issues there. 

See:
https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=mongocli

Note: Multiple mongocli versions seem to be marked with `includeInGlobalSearch : true`? within staging?


.drone.yml
```diff
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'DOCSP-32881'
```
casthewiz, line 11:
```suggestion
```

.drone.yml
```diff
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOCSP-32881'
```
casthewiz, line 11:
```suggestion
```

seungpark (commented):
the change LGTM but it also warrants unwanted changes. suggesting we tie in project name with the paths, ie.

```
resultMapping = {
  and: [{
    project: 'manual',
    slug: 'reference/operator/query/and'
  }, {
    project: 'manual',
    slug: 'reference/operator/aggregation/and'
  }] , ...
}
```

sarahemlin (approved):
Looks great!  Thank you so much for getting this in before the Paris Init meeting!

casthewiz:
@seungpark Agreed in spirit with the overlaps in slugs between projects being a problem - but given that the results for these terms are already pretty bad, I've talked with Sarah and we're going to merge this as an incremental improvement. 

I'll track in a separate ticket that we should look at using either `project` as a specifiable field or hook the mapping into fields in the taxonomy

seungpark:
> @seungpark Agreed in spirit with the overlaps in slugs between projects being a problem - but given that the results for these terms are already pretty bad, I've talked with Sarah and we're going to merge this as an incremental improvement.
> 
> I'll track in a separate ticket that we should look at using either `project` as a specifiable field or hook the mapping into fields in the taxonomy

separate ticket SGTM 👍

seungpark (approved):

PR #76, opened 2023-09-25
DOP-4048

seungpark:
[JIRA](https://jira.mongodb.org/browse/DOP-4048)

**Description:** This PR is intended to update the simplicity of our `/v2/status` route for the front end to easily render and use search facet filters with bare minimum of knowledge. The master list of our taxonomy will be represented as a list of [Facet Options](https://github.com/mongodb/docs-search-transport/pull/76/files#diff-2471679ba3ccc57dfd4dca806917d9d696f93bbc145984f7820ca4edfdcfc121R78-R85), which the front end can easily iterate over and decide between **group of checkboxes** and **checkbox filter**

[Current status response](https://docs-search-transport.mongodb.com/v2/status)
[Staged status response](https://docs-search-transport.docs.staging.corp.mongodb.com/v2/status)

Test cases have been updated to reflected the change

**Note:** for the Front End have responsive facet filters, it will have to use the `/v2/search/meta` route for refreshed facets that are filtered by each other. Follow up PR incoming. This current route is being refactored and not functioning, and this PR should not be merged until the route has been addressed. Specifics of changes are in first comment below V

### DIFF
--- a/src/Query/util.ts
+++ b/src/Query/util.ts
@@ -1,6 +1,6 @@
 import { Filter } from 'mongodb';
 
-import { Document, FacetAggregationStage, FacetDisplayNames, FacetOption } from '../SearchIndex/types';
+import { Document, FacetAggregationStage, FacetOption } from '../SearchIndex/types';
 
 const atomicPhraseMap: Record<string, string> = {
   ops: 'manager',
@@ -80,14 +80,23 @@ export const extractFacetFilters = (searchParams: URL['searchParams']): Filter<D
   return filter;
 };
 
-// TODO: update this to work with new facet structure (children and options)
 export const getFacetAggregationStages = (taxonomy: FacetOption[]) => {
   const facetKeysForAgg: FacetAggregationStage = {};
 
-  // TODO: traverse all of taxonomy to get facets for meta
-  for (const facetOption of taxonomy) {
-    console.log(facetOption);
+  function getKeysFromFacetOptions(facetOptions: FacetOption[]) {
+    for (const facetOption of facetOptions) {
+      facetKeysForAgg[facetOption.key] = {
+        type: 'string',
+        path: `facets.${facetOption.key}`,
+      };
+      for (const facetValue of facetOption.options) {
+        if (facetValue.facets?.length) {
+          getKeysFromFacetOptions(facetValue.facets);
+        }
+      }
+    }
   }
 
+  getKeysFromFacetOptions(taxonomy);
   return facetKeysForAgg;
 };
--- a/src/Query/util.ts
+++ b/src/Query/util.ts
@@ -1,6 +1,6 @@
 import { Filter } from 'mongodb';
 
-import { Document, FacetDisplayNames } from '../SearchIndex/types';
+import { Document, FacetAggregationStage, FacetDisplayNames, FacetOption } from '../SearchIndex/types';
 
 const atomicPhraseMap: Record<string, string> = {
   ops: 'manager',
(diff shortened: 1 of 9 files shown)


src/SearchIndex/index.ts
```diff
@@ -48,9 +48,13 @@ export class SearchIndex {
     const metaAggregationQuery = query.getMetaQuery(searchProperty, this.convertedTaxonomy);
     const cursor = this.documents.aggregate(metaAggregationQuery);
     try {
-      const aggRes = await cursor.toArray();
-      const res = formatFacetMetaResponse(aggRes[0] as FacetAggRes, this.convertedTaxonomy);
-      return res;
+      // TODO: re-implement
```
seungpark, line 51: Currently, the `/v2/search/meta` route will only shallow search the base level of facets and accept facet names to drill down.

The follow up PR will update this to be *expand all facets and accept filters for all levels of facets*. Return this in the format outlined in DOP-4048 for easy front end consumption

This is not required for the first pass at Search Facets UI (as they will be based off the `/v2/status/` route)

src/Query/util.ts
```diff
@@ -80,46 +80,14 @@ export const extractFacetFilters = (searchParams: URL['searchParams']): Filter<D
...
+  const facetKeysForAgg: FacetAggregationStage = {};
 
-  for (const baseFacet in taxonomy) {
-    facets[baseFacet] = {
-      type: 'string',
-      path: `facets.${baseFacet}`,
-    };
+  // TODO: traverse all of taxonomy to get facets for meta
+  for (const facetOption of taxonomy) {
+    console.log(facetOption);
```
rayangler, line 89: Do we want to keep this logged?

rayangler (commented):
Pretty much lgtm, just one comment about a log statement

src/SearchIndex/util.ts
```diff
@@ -32,6 +31,7 @@ interface FacetRes {
   [key: string]: FacetRes | string | number | undefined;
 }
 
+// TODO: update this to work with new facet structure (children and options)
 export function formatFacetMetaResponse(facetAggRes: FacetAggRes, taxonomyTrie: FacetDisplayNames) {
   const res: {
     count: number;
```
mmeigs, line 37: Nit: Should we comment out this function just to make sure no one tries to use it while it is not in sync with the new schema?

tests/resources/utils-data.ts
```diff
@@ -30,65 +30,270 @@ export const sampleTaxonomy = {
   ],
 } as Taxonomy;
 
-export const sampleFacetTrie = {
-  genres: {
+export const sampleFacetTrie = [
```
mmeigs, line 33: nitty nit nit: Should we rename this to sampleFacetOption?

mmeigs (approved):
Neat as a pin!

src/SearchIndex/util.ts
seungpark: in retrospect, i should have made this an additive PR but this is currently being used on front end for showing search results. can release together with follow up [PR](https://github.com/mongodb/docs-search-transport/pull/77) which fixes this route again

tests/resources/utils-data.ts
seungpark: good catch!! done. 

minor note, i will re add the `Trie` facet and the test. Trie structure and list structure are both stored server side for 1. validation against taxonomy and 2. taxonomy responses, respectively

src/Query/util.ts
seungpark: removed in follow up! https://github.com/mongodb/docs-search-transport/pull/77/files#diff-244d01de0824eda50531865572bc5dfaf1c3325fcb645aa6012177cfbdbb1017L87

src/SearchIndex/util.ts
mmeigs: ah fascinating, ok ok didn't realize!

PR #81, opened 2023-10-10
DOP-4071: allow either or/and for facet filters when performing search

seungpark:
[JIRA](https://jira.mongodb.org/browse/DOP-4071)

**Description:** PR provides an optional URL query param that allows users to dictate whether facets are combined using an OR/AND operator. Previously, all query params at different levels were treated as **AND** queries, and could therefore result in no results.

[Example of two filters passed and returning 0 results on prod.](https://docs-search-transport.mongodb.com/search?q=mongodb&facets.target_product=compass&facets.target_product%3Erealm%3Esub_product=web) Selected facets are `facets.target_product=compass` and `facets.target_product>realm>sub_product=web` but these are being treated as AND queries, and thus returning 0 results. ([see splunk log for aggregation query](https://mongodb.splunkcloud.com/en-US/app/search/show_source?sid=1696958328.580445&offset=12&latest_time=))

**Results:**

[Same query on branch returns results in stage (with default OR combination)
](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=mongodb&page=1&facets.target_product=compass&facets.target_product%3Erealm%3Esub_product=web)

[Splunk log showing OR combination
](https://mongodb.splunkcloud.com/en-US/app/search/show_source?sid=1696958694.581412&offset=2&latest_time=)

[Same query on branch shows no results when using AND combination
](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=mongodb&page=1&facets.target_product=compass&facets.target_product%3Erealm%3Esub_product=web&combineFilters=true)

[Splunk log showing AND combination](https://mongodb.splunkcloud.com/en-US/app/search/show_source?sid=1696958694.581412&offset=0&latest_time=)

With these changes, the default mechanism will be an "OR" combination of facets, and users have the option to combine them via URL query param `combineFilters`

### DIFF
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -15,7 +15,6 @@ export class Query {
   terms: Set<string>;
   phrases: string[];
   rawQuery: string;
-  // filters: Filter<Document>;
 
   /**
    * Create a new query.
@@ -26,7 +25,6 @@ export class Query {
     this.terms = new Set();
     this.phrases = [];
     this.rawQuery = queryString;
-    // this.filters = filters || {};
 
     const parts = queryString.split(/((?:\s+|^)"[^"]+"(?:\s+|$))/);
     let inQuotes = false;
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -15,18 +15,18 @@ export class Query {
   terms: Set<string>;
   phrases: string[];
   rawQuery: string;
-  filters: Filter<Document>;
+  // filters: Filter<Document>;
 
   /**
    * Create a new query.
    * @param {string} queryString The query to parse
    */
-  constructor(queryString: string, filters?: Filter<Document>) {
+  constructor(queryString: string) {
     console.log('Query parsing: ' + queryString);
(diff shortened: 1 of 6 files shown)


src/Query/index.ts
```diff
@@ -15,18 +15,18 @@ export class Query {
...
    * @param {string} queryString The query to parse
    */
-  constructor(queryString: string, filters?: Filter<Document>) {
+  constructor(queryString: string) {
     console.log('Query parsing: ' + queryString);
     this.terms = new Set();
     this.phrases = [];
     this.rawQuery = queryString;
-    this.filters = filters || {};
+    // this.filters = filters || {};
```
mmeigs, line 30: Should these be deleted? Or if left in, maybe leave a comment explaining?

mmeigs (approved):

src/Query/index.ts
seungpark: 🤦 thanks for the catch! removed

PR #82, opened 2023-10-13
DOP-4065: return search results' facets (in desired format)

seungpark:
[JIRA](https://jira.mongodb.org/browse/DOP-4065)

See second bullet point for A/C: `/search route returns documents with "facets" field`

This PR is intended to return the `document.facets` fields in a similar format as the [/v2/status response](https://docs-search-transport.docs.staging.corp.mongodb.com/v2/status), namely with facet values having `{key, id}` properties. This allows for front end to map this to an existing facet from the above response.

[Current search results do not return `facets` field](https://docs-search-transport.mongodb.com/search?q=meta)

[Staging branch returns `facets` field, along with all previous fields in the above example](https://docs-search-transport.docs.staging.corp.mongodb.com/search?q=meta)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOP-4065'
 
 steps:
   - name: publish-staging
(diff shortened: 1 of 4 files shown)


.drone.yml
```diff
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOP-4065'
```
seungpark, line 11:
```suggestion
```

reminder to revert before merge

mmeigs (approved):
LGTM. Thanks for the comments!!

rayangler (approved):

PR #84, opened 2023-11-01
DOP-4089: treat same level selections as "or"

seungpark:
DOP-4089 discussions (on [slack](https://mongodb.slack.com/archives/GGPUBUMLN/p1698844414318519) and [tech spec](https://docs.google.com/document/d/1c9P5_wnlLswoZftJbZG-baqG1Fgs26lO1gv7ExgT3Ss/edit#heading=h.ytznsphw0fe2) included allowing same base level selections as "OR" statements, while treating selections of different base categories as "AND" statements.

This PR aims to do this, by combining query params of the **same base facet**, into a nested compound SHOULD operator, under the root MUST compound operator.

As an example:

Query route:
```
/search/ ?q=$and
&facets.genre=tutorial
&facets.target_product>atlas>sub_product=search
&facets.target_product=drivers
``` 
previously, this translated to MUST statements for each line. and thus, it would return 0 results, since all documents are tagged with either `atlas` or `driver` target product facet, (there is no overlap).
Aggregation translation from above query params:
```
"compound": {
  "must": [
     ....
     {
            "phrase": {
              "query": "tutorial",
              "path": "facets.genre"
            }
      },
     {
            "phrase": {
              "query": "search",
              "path": "facets.target_product>atlas>sub_product"
            }
      },
     {
            "phrase": {
              "query": "drivers",
              "path": "facets.target_product"
            }
      ],
} 
```

This PR translates the same request parameters to the following aggregation:
```
{
  "compound": {
    "must": [
      ...
      {                 <-- compound below translates to "MUST have at least one of these should compounds for genre"
        "compound": {
          "should": [
            { "phrase": { "query": "tutorial", "path": "facets.genre" } }
          ]
        }
      },
      {               <-- compound below translates to "MUST have at least one of these should compounds for target_product"
        "compound": {
          "should": [
            {
              "phrase": {
                "query": "search",
                "path": "facets.target_product>atlas>sub_product"
              }
            },
            {
              "phrase": {
                "query": "drivers",
                "path": "facets.target_product"
              }
            }
          ]
        }
      }
    ]
  }
}
```

### DIFF
--- a/src/Query/util.ts
+++ b/src/Query/util.ts
@@ -53,14 +53,24 @@ export const extractFacetFilters = (searchParams: URL['searchParams']): Filter<D
   // where each query param starting with 'facets.' denotes a filter
   const FACET_PREFIX = 'facets.';
   const filters: Filter<Document>[] = [];
+
+  // values with same keys should be treated as OR
+  // store in hash with one pass and another pass of the hash to filters
+  const queryParamLists: { [key: string]: string[] } = {};
+
   for (const [key, value] of searchParams) {
     if (!key.startsWith(FACET_PREFIX)) {
       continue;
     }
 
+    queryParamLists[key] = queryParamLists[key] || [];
+    queryParamLists[key].push(value);
+  }
+
+  for (const [key, values] of Object.entries(queryParamLists)) {
     filters.push({
       phrase: {
-        query: value,
+        query: values,
         path: key,
       },
     });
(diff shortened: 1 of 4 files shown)


src/Query/util.ts
```diff
@@ -1,6 +1,6 @@
 import { Filter } from 'mongodb';
 
-import { Document, FacetAggregationStage, FacetOption } from '../SearchIndex/types';
+import { Document, FacetAggregationStage, FacetOption, TrieFacet } from '../SearchIndex/types';
```
rayangler, line 3: (Nit) Looks like `TrieFacet` is unused

rayangler (approved):
Lgtm, thanks for the breakdown in the PR description

src/Query/util.ts
seungpark: good catch! removed

PR #86, opened 2023-11-08
DOP-3984 added analyzers for synonyms

anabellabuckvar:
### Tickets
[DOP-3984](https://jira.mongodb.org/browse/DOP-3984)

### Current behavior
Currently, there is a synonym file within the docs-search-transport repository that maintains synonyms, but synonyms are not utilized in search queries.

The results for the below search queries, "js" and "javascript" on the prod docs site are different.
https://www.mongodb.com/docs/search/?q=js&page=1
https://www.mongodb.com/docs/search/?q=javascript&page=1

### Notes
Testing was primarily conducted by comparing document results and number of results between the following two aggregations in `search-staging.documents`. Query and path fields were changed to test for different queries in different fields

Screenshot 2023-11-09 at 2 23 34 PM
Screenshot 2023-11-09 at 2 23 59 PM

This PR added support for synonyms in search queries by specifying `synoymns` in the $search aggregation such that all synonym queries use the `lucene.whitespace` analyzer. Although there is now some functionality for synonyms, it is limited because of several issues:

`lucene.whitespace` analyzer only tokenizes on whitespace, and does not use a capitalization filter to normalize search queries or document fields to lowercase. Fixing this would likely require a new custom analyzer.
 For example: "admin" and "administration" are synonyms. Therefore, searchQuery 'admin' returns results for 'administration', 'admin, and 'Admin'(this only because the default analyzer for most fields is `lucene.english` or `lucene.standard` which do normalize to lowercase). However, it does not return any results for the capitalization of its synonym, "Administration".


src/data/atlas-search-index.ts
```diff
@@ -98,7 +126,7 @@ export const SearchIndex: IndexMappings = {
       source: {
         collection: SYNONYM_COLLECTION_NAME,
       },
-      analyzer: 'lucene.keyword',
+      analyzer: 'lucene.whitespace',
```
branberry, line 129: If we think it's worthwhile, we could make a custom analyzer to address the issue of capitalization with the whitespace analyzer.

Here's what it could look like:

```json
"analyzers": [
  {
    "name": "synonymAnalyzer",
    "tokenizer": {
      "type": "whitespace"
    },
    "tokenFilters": [
         {
             "type": "lowercase" 
         }
     ]
  }
]
```

You should be able to reference this custom analyzer here with the `name` property.

[custom analyzer reference](https://www.mongodb.com/docs/atlas/atlas-search/analyzers/custom/)
[lowercase token filter reference](https://www.mongodb.com/docs/atlas/atlas-search/analyzers/token-filters/#lowercase)

branberry (commented):
Hey Anabella! Thank you for the detailed write-up. I added one comment about creating a custom analyzer to address the capitalization issue mentioned in the PR.

src/data/atlas-search-index.ts
seungpark: great call @branberry ! we can also expand on this analyzer in the future (ie. when adding multi whitespace words)

+1 on this suggestion

anabellabuckvar: Great idea, thanks @branberry!! I updated it and it fixed the capitalization issue. 

Also @seungpark, multi whitespace word synonyms are already working so that ~shouldn't be necessary :)

branberry (commented):
Hey Anabella! Just have one more small nit, otherwise this looks great!

branberry (approved):
LGTM! Nice work on this one

seungpark (approved):
this LGTM as well! 🚀

PR #90, opened 2023-11-27
DOP-4121: Boost/bury based on product usage & pageviews

anabellabuckvar:
### Ticket
[DOP-4121](https://jira.mongodb.org/browse/DOP-4121)

### Context
Realm docs are currently overrepresented in search result sets by a slight margin. This PR buries results with the `searchProperty` `realm-master` but keeps all other results, as well as their order, identical. The approach used was the suggested one for burying/boosting certain results as detailed [here](https://docs.google.com/document/d/1G1RYHBbFXqm5kG98GETCTvKj1UJHv5rzDHG18X-5jYM/edit#heading=h.sq6vzuky8sas)

## NOTES

- Testing: Only a limited amount of testing could be done. To test, query results from the local staging instance were compared with the results of the same queries on the staging server. 
Ex: https://docs-search-transport.docs.staging.corp.mongodb.com/search/?q=sdk compared to http://localhost:8080/search/?q=sdk. Several queries were tested, and the behavioral outcomes seemed to be as desired. Queries that do not yield any 'Realm docs' results on the staging server yielded the the same results in the same order on the local staging instance.  Queries that do yield some 'Realm docs' results on the staging server yielded results that were identical except for that docs with 'Realm-master' search property were lower down (buried) in the search results. However, it should be noted that it was not possible to check the score of any of the result documents and so it couldn't be verified whether they were buried exactly, and only, to the desired extent.

- "`Any`" typing is frequent in `Query/index.ts`
- The ticket asks if we should always bury, e.g. will there be realm-specific search terms that we should ignore? This has not been discussed further

### DIFF
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -10,6 +10,43 @@ function processPart(part: string): string[] {
   return tokenize(part, false);
 }
 
+//check type of parts
+function constructAgg(parts: any[]): object[] {
+  console.log("THE PARTS", parts);
+  const newParts: any[] = [];
+  for(var part of parts){ 
+    newParts.push({ compound: {
+      must: [part],
+      mustNot: [
+        {
+          text: {
+            query: 'Realm',
+            path: ['searchProperty']
+          }
+        }
+      ]
+     } 
+    },
+    //bury result where query matched clause AND in Realm docs
+      { compound: {
+      must: [
+        part,
+        {
+          text:{
+            query: 'Realm',
+            path: ['searchProperty']
+          }
+        }
+      ],
+      score: {"boost": {value: 0.5}}
+     } 
+    });
+    //push to the two compounds as one should clause to the array
+  }
+  console.log(newParts[0].compound, newParts[0].compound.mustNot, newParts[0].compound.must[0]);
+  return newParts;
+}
+
 /** A parsed search query. */
 export class Query {
   terms: Set<string>;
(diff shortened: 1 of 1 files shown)


src/Query/index.ts
```diff
@@ -10,6 +10,49 @@ function processPart(part: string): string[] {
...
+    newParts.push(
+      //if given query matches a "part" result not in BURIED_PROPERTY(ex: Realm) docs, score remains unaffected
+      {
+        compound: {
+          must: [part],
+          mustNot: [
+            {
+              text: {
+                query: BURIED_PROPERTIES,
+                path: 'searchProperty',
```
rayangler, line 30: Just curious: I'm assuming `searchProperty` is easier to use (and more convenient right now) than using the `facets` field?

src/Query/index.ts
```diff
@@ -10,6 +10,49 @@ function processPart(part: string): string[] {
   return tokenize(part, false);
 }
 
+const BURIED_PROPERTIES = ['realm'];
+const BURIED_FACTOR = 0.8;
+
+//check type of parts
+function constructAgg(parts: any[]): object[] {
+  const newParts: any[] = [];
+  for (var part of parts) {
```
rayangler, line 19: (Nit) Can we use `let` or `const` instead of `var`? Just want to avoid unnecessary hoisting

rayangler (commented):
Not an expert on Atlas Search stuff, but code looks generally good to me with one nit. 

Tested it by burying properties other than Realm, and it seems to help lower their placements in the search results list! Thank you for the helpful comments

src/Query/index.ts
anabellabuckvar: Yes, also because most search-staging documents have a `null` `facets` field

seungpark (approved):
this LGTM. great work on burying! tested that it works with v1 and v2 search.

rayangler (approved):

PR #92, opened 2023-11-30
DOP-4145: Add context to Docs Search Transport errors

anabellabuckvar:
### Ticket
[DOP-4145](https://jira.mongodb.org/browse/DOP-4145)

### Currrent Behavior
Docs Search Transport [Splunk logs](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22docs-prod%22%20%20source%3D%22*search-transport*%22%20%5C(error%5C)&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1700024400&latest=1700715600&sid=1700685979.1211656) show several errors. The two that are the focus of this ticket are URL Type Errors and facet fetching errors. Examples of each are included below
Screenshot 2023-11-30 at 2 04 02 PM
Screenshot 2023-11-30 at 2 06 29 PM

Error handling and logging of such errors is now improved to provide greater context for debugging purposes

### DIFF
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -64,10 +64,10 @@ export default class Marian {
     try {
       parsedUrl = new URL(url, `http://${req.headers?.host}`);
     } catch (e) {
-log.warn(`URL constructor could not create a URL with url ${url} and base ${req.headers?.host}`);
-res.writeHead(500, {});
-res.end('');
-return;
+      log.warn(`URL constructor could not create a URL with url ${url} and base ${req.headers?.host}`);
+      res.writeHead(500, {});
+      res.end('');
+      return;
     }
     const pathname = (parsedUrl.pathname || '').replace(/\/+$/, '');
     if (pathname === '/search') {
@@ -265,7 +265,7 @@ return;
       const toml = await res.text();
       return parse(toml);
     } catch (e) {
-      console.error(`Error while fetching taxonomy: ${JSON.stringify(e)}`);
+      log.error(`Error while fetching taxonomy: ${JSON.stringify(e)}`);
       throw e;
     }
   }
@@ -301,7 +301,7 @@ return;
       res.facets = sortFacets(res.facets);
       return res;
     } catch (e) {
-      console.error(
+      log.error(
         `Error fetching facet metadata for query ${rawQuery}, with search property ${searchProperty}, and filters ${filters}. ${JSON.stringify(
           e
         )}`
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -63,8 +63,7 @@ export default class Marian {
     let parsedUrl: URL;
     try {
       parsedUrl = new URL(url, `http://${req.headers?.host}`);
-    }
-    catch (e){
+    } catch (e) {
       console.error(`URL constructor could not create a URL with url ${url} and base ${req.headers?.host}`);
       console.trace();
       throw e;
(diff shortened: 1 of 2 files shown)


src/SearchIndex/index.ts
```diff
@@ -72,15 +72,19 @@ export class SearchIndex {
       const aggRes = await cursor.toArray();
       return formatFacetMetaResponse(aggRes[0] as FacetAggRes, this.trieFacets);
     } catch (e) {
-      log.error(`Error while fetching facets: ${JSON.stringify(e)}`);
+      log.error(
+        `Error while fetching facets for query ${query}, with search property ${searchProperty}, and filters ${filters} ${JSON.stringify(
```
rayangler, line 76: Testing locally, it seems like we just print `[object Object]` when printing out objects like `query`. I think it might make sense to use `rawQuery` or stringify any objects we expect to print

src/Marian/index.ts
```diff
@@ -292,12 +297,31 @@ export default class Marian {
...
+      console.trace();
+      throw e;
+    }
     try {
-      const res = await this.index.fetchFacets(query, searchProperty, filters);
       res.facets = sortFacets(res.facets);
       return res;
     } catch (e) {
-      console.error(`Error fetching facet metadata: ${JSON.stringify(e)}`);
+      console.error(
```
rayangler, line 319: I'm not sure if having 2 separate try-catch blocks here is necessary since they're both only used to log and re-throw the error. In other words, we're not necessarily handling these two errors differently (nor do we need to right now).

My recommendation would be to re-consolidate the two, but keep the context on the arguments used if anything. As long as we have context for the error that occurred when fetching metadata, it should allow us to reproduce locally to debug.

src/Marian/index.ts
```diff
@@ -60,10 +60,15 @@ export default class Marian {
...
     }
-    const parsedUrl = new URL(url, `http://${req.headers.host}`);
-
+    let parsedUrl: URL;
+    try {
+      parsedUrl = new URL(url, `http://${req.headers?.host}`);
+    } catch (e) {
+      console.error(`URL constructor could not create a URL with url ${url} and base ${req.headers?.host}`);
+      console.trace();
+      throw e;
```
rayangler, line 69: I was thinking about this error, and I'm wondering if we should just avoid logging an error (or maybe just log a warning) and send an error as a response, instead of throwing.

Otherwise, I'm afraid we'll keep logging noisy errors for parsing urls from requests we can't control. Any thoughts?

anabellabuckvar: Sending an error as a response instead makes a lot of sense. For the purpose of investigating the erroring requests right now, I think  it would be helpful to also log a warning with information on bad requests

src/Marian/index.ts
```diff
@@ -291,13 +295,17 @@ export default class Marian {
...
       searchProperty = [searchProperty];
     }
-
     try {
       const res = await this.index.fetchFacets(query, searchProperty, filters);
       res.facets = sortFacets(res.facets);
       return res;
     } catch (e) {
-      console.error(`Error fetching facet metadata: ${JSON.stringify(e)}`);
+      console.error(
```
rayangler, line 303: (Nit) We can probably use `log.error` instead of `console.error` to help consolidate the log formatting

rayangler (approved):

PR #93, opened 2023-12-04
[DOP-4164]: Resolve Splunk Alert for "empty file: search-indexes/prd/"

branberry:
### Ticket
[DOP-4164](https://jira.mongodb.org/browse/DOP-4164)

### Notes

[StackOverflow answer as to why this happens](https://stackoverflow.com/a/75620490/6765567). Basically, if someone creates a folder in an S3 bucket from the AWS console, it will create an object for it. Instead of warning about an empty file, we can ignore objects with a size of 0 and/or check if its a directory. Because of this, we can safely remove the `console.error` statement.

Also, the memory request and limit amount has been increased to (hopefully) resolve the out of memory errors that we have been receiving. Based on data from Kanopy, peak memory usage is around 1100 Mi, which is what our memory request threshold is at. My hypothesis is that we hit this threshold-which is how much memory is guaranteed to be allocated-and then receive an out of memory error. The pod still runs, but it looks like the application is a bit hosed up, so adding more memory should be enough to resolve this issue.

image


src/SearchIndex/util.ts
```diff
@@ -462,7 +462,6 @@ async function getManifestsFromS3(bucketName: string, prefix: string): Promise<M
   const manifests = [];
   for (const bucketEntry of result.Contents || []) {
     if (bucketEntry.Size === 0) {
```
seungpark, line 464: can we actually validate its not a "directory file" ie doesnt end in a trailing slash, and log it if so?

seungpark (changes requested):
great find! thanks for the supplementary post as well

seungpark (approved):
❤️

PR #99, opened 2024-03-13
DOP-4182: Specify types for query builder in search transport

mayaraman19:
### Ticket

DOP-4182

### Notes

I tried to update `any` types when it seemed pretty straightforward. Here are some places where I did not change types (as well as left a couple other comments detailing reasoning):
- [src/data/atlas-types.ts](https://github.com/mongodb/docs-search-transport/blob/main/src/data/atlas-types.ts#L9) - I couldn't really easily find enough information on the returned `SearchIndexResponse` to know what types should exist in `customAnalyzer`. 
- [src/Marian/index.ts](https://github.com/mongodb/docs-search-transport/blob/main/src/Marian/index.ts#L135) - don't feel it's necessary to type the error
- [src/SearchIndex/util.ts](https://github.com/mongodb/docs-search-transport/blob/main/src/SearchIndex/util.ts#L176) - not sure what the `any` in `Record<string,any>` would look like, figured at least the typing added will suffice

This PR also updates the README with some testing info, as well as adds checks for the `TAXONOMY_URL` environment variable.

### README updates

### DIFF
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -3,11 +3,12 @@ import assert from 'assert';
 import Logger from 'basic-logger';
 import http from 'http';
 import { parse } from 'toml';
+import { Document } from 'mongodb';
 
 import { checkAllowedOrigin, checkMethod } from './util';
 import { StatusResponse } from './types';
 import { SearchIndex } from '../SearchIndex';
-import { Taxonomy } from '../SearchIndex/types';
+import { FacetMeta, Taxonomy } from '../SearchIndex/types';
 import { AtlasAdminManager } from '../AtlasAdmin';
 import { setPropertyMapping } from '../SearchPropertyMapping';
 import { Query, InvalidQuery } from '../Query';
@@ -233,7 +234,7 @@ export default class Marian {
     res.end(responseBody);
   }
 
-  private async fetchResults(parsedUrl: URL): Promise<any[]> {
+  private async fetchResults(parsedUrl: URL): Promise<Document[]> {
     const rawQuery = (parsedUrl.searchParams.get('q') || '').toString();
 
     if (!rawQuery) {
@@ -283,7 +284,7 @@ export default class Marian {
     res.end(responseBody);
   }
 
-  private async fetchFacetMeta(parsedUrl: URL): Promise<any> {
+  private async fetchFacetMeta(parsedUrl: URL): Promise<FacetMeta> {
     const rawQuery = (parsedUrl.searchParams.get('q') || '').toString();
     if (!rawQuery || !rawQuery.length) {
       throw new InvalidQuery();
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -1,4 +1,4 @@
-import { Filter } from 'mongodb';
+import { Filter, Document as mdbDocument } from 'mongodb';
 import { getFacetAggregationStages, getProjectionAndFormatStages, tokenize } from './util';
 import { Document, FacetOption } from '../SearchIndex/types';
 import { getPropertyMapping } from '../SearchPropertyMapping';
@@ -103,7 +103,7 @@ export class Query {
     }
   }
 
-  getCompound(searchProperty: string[] | null, filters: Filter<Document>[]) {
+  getCompound(searchProperty: string[] | null, filters: Filter<Document>[]): Compound {
     const terms = Array.from(this.terms);
     const parts: Part[] = [];
     const searchPropertyMapping = getPropertyMapping();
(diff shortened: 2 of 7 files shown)


src/Marian/index.ts
```diff
@@ -233,7 +234,7 @@ export default class Marian {
     res.end(responseBody);
   }
 
-  private async fetchResults(parsedUrl: URL): Promise<any[]> {
+  private async fetchResults(parsedUrl: URL): Promise<Document[]> {
```
mayaraman19, line 237: because the results returned are an aggregation, I don't think there's any way to be sure of the type except for the fact that it will be a `mongodb.Document` type.

src/Query/index.ts
```diff
@@ -179,7 +180,7 @@ export class Query {
       },
     });
 
-    const compound: { should: any[]; must: any[]; filter?: any[]; minimumShouldMatch: number } = {
+    const compound: { should: CompoundPart[]; must: Must[]; filter?: any[]; minimumShouldMatch: number } = {
```
mayaraman19, line 183: i left the `filter` parameter as `any[]` because I could not see where it is used.

anabellabuckvar (approved):
Thanks for doing this!

src/Query/index.ts
```diff
@@ -256,7 +257,7 @@ export class Query {
     return agg;
   }
 
-  getAggregationQuery(searchProperty: string[] | null, filters: Filter<Document>[], page?: number): any[] {
+  getAggregationQuery(searchProperty: string[] | null, filters: Filter<Document>[], page?: number): Filter<Document>[] {
```
seungpark, line 260: same as noted below, `collection.aggregate` expects a Document[] type to be passed, as opposed to Filter<Document> 

https://github.com/mongodb/node-mongodb-native/blob/main/src/collection.ts#L936

src/index.ts
```diff
@@ -59,7 +62,9 @@ function verifyAndGetEnvVars() {
     if (!adminPubKey) {
       console.error(`Missing ${ADMIN_PUB_KEY}`);
     }
-    // TODO: add taxonomy url
+    if (!taxonomyUrl) {
```
seungpark, line 65: ❤️

src/SearchIndex/util.ts
```diff
@@ -167,7 +169,7 @@ export function convertTaxonomyToTrie(taxonomy: Taxonomy): TrieFacet {
     name: '',
   };
 
-  function addToRes(entityList: TaxonomyEntity[], ref: { [key: string]: any }, property: string) {
+  function addToRes(entityList: TaxonomyEntity[], ref: RefObj, property: string) {
```
seungpark, line 172: think we can replace RefObj with `TrieFacet` and refer to re-assigned `ref` `as Triefacet` and have the entity declared as one as well `entity: Triefacet = {}`

seungpark (changes requested):
thank you for doing this!! left some feedback regarding `Filters` and a suggestion for the ambiguous recursive reference object. PTAL!

seungpark:
as for the three highlighted in the comment:

- we do not use these tokenFilters or charFilters in the custom analyzer field so i'm fine with having these until we use them ( can add a comment or comment these fields out)
- that is a limitation of typescript for purposely leaving the catch block as unknown. we have to typecheck ie. `err instanceof Error`, `message in err`
- [my comment](https://github.com/mongodb/docs-search-transport/pull/99#discussion_r1524953588) for re-using TrieFacet should make the any go away here

src/Query/types.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+type Equals = {
+  equals: {
+    path: string | string[];
+    value: boolean;
+  };
+};
+
+export type Must = Phrase | Equals | Filter<Document>;
```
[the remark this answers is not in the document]
mayaraman19: I included this because [here](https://github.com/mongodb/docs-search-transport/blob/main/src/Query/index.ts#L234) we concatenate `filters` with `must` and `filters` is of type `Filter<Document>[]`.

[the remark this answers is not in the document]
mayaraman19: should I change it to `Document`?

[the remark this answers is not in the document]
seungpark: mongodb.Document makes sense here. and the Filter can still pass 👍

src/Query/index.ts
seungpark: to be clear, a Mdb.Document 😅

src/Query/index.ts
```diff
@@ -256,7 +257,7 @@ export class Query {
     return agg;
```
seungpark, line 257: minor, can we specify return type for this fn to be mdb.Document[] as well. used for `aggregate(agg)` function

src/Query/index.ts
mayaraman19: haha was just about to ask. seems like all in this file the `Filter` type is being cast to a `SearchIndex/types.ts/Document` type, should this all be changed to `mongodb.Document`?

seungpark (approved):
LGTM! thanks for doing this !

PR #100, opened 2024-03-25
DOP-3976: add a boost for phrase of the original query

seungpark:
### Ticket

DOP-3976

### Notes
- Adds a [phrase](https://mongodb.com/docs/atlas/atlas-search/phrase/) operator to search across documents' `paragraphs` `headings` and `text` for the original search phrase with standard analyzer. Added a boost to this score as we would want documents with these words in order to be higher in the ranking

### Staged changes:
Can see the change here:
https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/search/?q=max+disk+iops&page=1

### README updates

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOP-3976'
 
 steps:
   - name: publish-staging
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,6 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
-      - 'refs/heads/DOP-3976'
 
 steps:
   - name: publish-staging
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -195,7 +195,7 @@ export class Query {
           query: terms.join(' '),
           score: {
             boost: {
-              value: 50,
+              value: 10,
             },
           },
         },
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -187,6 +187,19 @@ export class Query {
     };
     const searchPropertyNames = Object.keys(searchPropertyMapping);
 
+    // DOP-3976: phrases found in conjunction should have additional score boost
+    compound.should.push({
+      phrase: {
+        path: ['paragraphs', 'text', 'headings'],
+        query: terms.join(' '),
+        score: {
+          boost: {
+            value: 50,
+          },
+        },
+      },
+    });
+
     // if user requested searchProperty, must match this property name
     // allowing mix usage of searchProperty and facets
     if (searchProperty !== null && searchProperty.length !== 0) {
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -189,13 +189,15 @@ export class Query {
 
     // DOP-3976: phrases found in conjunction should have additional score boost if they are found in order
     if (terms.length > 1) {
+      const maxLength: number = terms.reduce((max, term) => Math.max(max, term.length), 0);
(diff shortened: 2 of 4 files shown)


.drone.yml
```diff
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOP-3976'
```
mayaraman19, line 11: Should this line be deleted?

src/Query/index.ts
```diff
@@ -187,6 +187,21 @@ export class Query {
     };
     const searchPropertyNames = Object.keys(searchPropertyMapping);
 
+    // DOP-3976: phrases found in conjunction should have additional score boost if they are found in order
```
mayaraman19, line 190: Should we add a `slop` value to this operator? Currently on the staging link, the results for, for example, "delete collection" and "collection delete" are quite different. Would it be worth adding a little bit of a `slop` value so that there's a bit of flexibility in how the query is asked?

tests/unit/Query.test.ts
```diff
@@ -31,4 +32,16 @@ describe('Query', () => {
     deepStrictEqual(query.terms, new Set(['officially', 'supported']));
     deepStrictEqual(query.phrases, ['officially supported']);
   });
+
+  it('should look for the phrase as a compound should with a boost', () => {
```
mayaraman19, line 36: Could this test name be reworded? It took me a bit of time to understand what it means 😭 the "should" threw me off lol

.drone.yml
seungpark: will delete before merge!

tests/unit/Query.test.ts
seungpark: good point! updated

src/Query/index.ts
seungpark: this can be a follow up but i think this will introduce more changes than the ask, in terms of ranking. this is specifically for the phrase asked without reordering.

but you bring up a good point - perhaps the boost is too large, as it will look for this phrase and boost for its existence - regardless of its content. will lower this boost while trying to keep the same result as described in the ticket

seungpark: @mayaraman19 added a slop factor for this and modified the boost! upon testing, this was pretty significant in boosting the phrase as is, and also preserving cases where order does not matter as in your example.

mmeigs (commented):
This looks good. But I am wondering about the results of the search. Mainly how closely the phrase is found. I'm sure you've sent this to Sarah, but I did notice that the one result she said should be found by this search phrase is `https://www.mongodb.com/docs/atlas/review-available-metrics/` and that is currently on the second page of results, the 16th search result.

seungpark:
> This looks good. But I am wondering about the results of the search. Mainly how closely the phrase is found. I'm sure you've sent this to Sarah, but I did notice that the one result she said should be found by this search phrase is `https://www.mongodb.com/docs/atlas/review-available-metrics/` and that is currently on the second page of results, the 16th search result.

good point. i think its safe to raise this boost factor since we added a slop factor. also left [a comment](https://jira.mongodb.org/browse/DOP-3976?focusedCommentId=6214136&focusedId=6214136&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-6214136) for possible difference in result on the issue

mayaraman19 (approved):
I think this looks good, I definitely see an improvement with the other examples in the ticket such as "delete collection" and "node.js driver" (and with slop factor, "collection delete" and "driver node.js"). Might want to get the results approved by Sarah as well when she's back?

sarahemlin (approved):
LGTM from Corry Root & me.  Thank you!

PR #101, opened 2024-03-29
DOP-4489: supply list of manifest files in S3

seungpark:
### Ticket

DOP-4489

This change opens up a new route `/manifests` that provides a link of JSON files that smartling's crawlers can access, for pre processing and translations.

### Notes
- Previously, we were setting the full URI for the manifest s3 location, so added a few more env variables without changing the existing logic.
- This does not require any changes to the README but additional changes have been made to env variables (sample env, drone env variables)

### README updates

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOP-4489'
 
 steps:
   - name: publish-staging
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,6 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
-      - 'refs/heads/DOP-4489'
 
 steps:
   - name: publish-staging
--- a/sample.env
+++ b/sample.env
@@ -1,5 +1,7 @@
 # URI in s3 for source manifests
 MANIFEST_URI='s3://docs-search-indexes-test/search-indexes/preprd'
+S3_BUCKET= docs-search-indexes-test
+S3_PATH: search-indexes/preprd
 # Database containing indexed search results
 ATLAS_DATABASE='search-staging'
 # Atlas connection string uri
--- a/sample.env
+++ b/sample.env
@@ -1,7 +1,7 @@
 # URI in s3 for source manifests
 MANIFEST_URI='s3://docs-search-indexes-test/search-indexes/preprd'
-S3_BUCKET= docs-search-indexes-test
-S3_PATH: search-indexes/preprd
+S3_BUCKET=docs-search-indexes-test
+S3_PATH=search-indexes/preprd
 # Database containing indexed search results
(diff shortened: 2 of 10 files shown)


sample.env
```diff
@@ -1,5 +1,7 @@
 # URI in s3 for source manifests
 MANIFEST_URI='s3://docs-search-indexes-test/search-indexes/preprd'
+S3_BUCKET= docs-search-indexes-test
+S3_PATH: search-indexes/preprd
```
rayangler, line 4: Should this be `S3_PATH=`?

src/index.ts
```diff
@@ -40,16 +45,24 @@ The following environment variables are used:
...
-  if (!manifestUri || !atlasUri || !groupId || !adminPrivKey || !adminPubKey) {
+  if (!manifestUri || !s3Bucket || !atlasUri || !groupId || !adminPrivKey || !adminPubKey || !s3Path) {
     if (!manifestUri) {
       console.error(`Missing ${MANIFEST_URI_KEY}`);
     }
+    if (!s3Bucket) {
+      console.error(`Missing ${S3_BUCKET_KEY}`);
+    }
+    if (!s3Path) {
+      console.error(`Missing ${S3_BUCKET_KEY}`);
```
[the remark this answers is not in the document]
seungpark: good catch!

src/index.ts
```diff
@@ -40,16 +45,24 @@ The following environment variables are used:
 
 function verifyAndGetEnvVars() {
   const manifestUri = process.env[MANIFEST_URI_KEY];
+  const s3Bucket = process.env[S3_BUCKET_KEY];
+  const s3Path = process.env[S3_PATH_KEY];
```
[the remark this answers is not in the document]
seungpark: i was thinking about this first, but we also [allow MANIFEST_URI to be a directory as well ](https://github.com/mongodb/docs-search-transport/blob/main/src/SearchIndex/util.ts#L554) and in this case, we would already need to provide a separate bucket name and path as well

sample.env
seungpark: good catch! updated

src/index.ts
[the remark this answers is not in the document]
rayangler: Oh, interesting. Thanks for pointing out that use case. I think it's awkward either way since if a local directory is being used as the URI, the new endpoint will return a list of S3 URLs that may not even be reflective of the manifests in the directory being used. For example, manifests in the directory can have manifests that don't exist in the S3 bucket, and vice-versa.

That being said, this isn't a big deal and was just a small concern. I'm okay with keeping as-is

.drone.yml
```diff
@@ -8,6 +8,7 @@ trigger:
     include:
       - 'refs/tags/v*'
       - 'refs/heads/main'
+      - 'refs/heads/DOP-4489'
```
rayangler, line 11: Formal reminder to remove before merging, just in case

rayangler (approved):

.drone.yml
seungpark: done

PR #102, opened 2024-04-03
DOP-4503: check that property mapping is only checking for string properties

seungpark:
### Ticket

DOP-4503

### Notes
- See ticket for explanation of bug. [In the case shown in this log](https://mongodb.splunkcloud.com/en-US/app/search/show_source?sid=scheduler_cmF5bXVuZC5yb2RyaWd1ZXo__search__RMD5f5aaa11866325be7_at_1712116800_3520&offset=0&latest_time=), user is searching for `constructor` which exists as a property on the object, but does not exist in the constant mapping as a key. The fix here is to check that the property exists as a type of `String[]`
- Added this case as a unit test to prevent future errors of the same kind.

### README updates

### DIFF
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -109,7 +109,9 @@ export class Query {
     const searchPropertyMapping = getPropertyMapping();
 
     // if we need to boost for matching slug on an exact rawQuery match
-    if (strippedMapping[this.rawQuery.trim()]) {
+    const boostedStrings = strippedMapping[this.rawQuery.trim()];
+    if (Array.isArray(boostedStrings) && typeof boostedStrings[0] === 'string') {
+      // if ( strippedMapping[this.rawQuery.trim()]) {
       parts.push({
         text: {
           path: 'strippedSlug',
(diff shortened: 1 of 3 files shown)


src/Query/index.ts
```diff
@@ -109,7 +109,8 @@ export class Query {
     const searchPropertyMapping = getPropertyMapping();
 
     // if we need to boost for matching slug on an exact rawQuery match
-    if (strippedMapping[this.rawQuery.trim()]) {
+    const boostedStrings = strippedMapping[this.rawQuery.trim()];
+    if (Array.isArray(boostedStrings) && typeof boostedStrings[0] === 'string') {
       parts.push({
         text: {
           path: 'strippedSlug',
```
mmeigs, line 116: You can also continue to use `boostedStrings` a few lines down rather than trim again. 

Also, I'm not as familiar with the intricacies of this. Is it enough to check the first element's type because there's no other combination of types in an array this could be, or should we go the extra mile and check all the array's element's types. Not blocking at all, just me wondering.

mmeigs (commented):
Thanks for the test and the logging!

src/Query/index.ts
seungpark: good catch! done.

i think its valid to check since we have the type defined as 
```
interface ResultMapping {
  [key: string]: string[];
}
```

rayangler (approved):

PR #103, opened 2024-04-04
DOP-3934: prereq - dev server for 

seungpark:
### Ticket

DOP-3934

### Notes
- Update readme to include instructions for deployment dev

### README updates

- - [X] This PR introduces changes that should be reflected in the README, and I have made those updates.
- - [ ] This PR does not introduce changes that should be reflected in the README

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -6,9 +6,9 @@ name: default
 trigger:
   ref:
     include:
-      - 'refs/tags/v*'
-      - 'refs/heads/main'
-      - 'refs/heads/DOP-*'
+      - "refs/tags/v*"
+      - "refs/heads/main"
+      - "refs/heads/DOP-*"
 
 steps:
   - name: publish-dev
@@ -28,7 +28,7 @@ steps:
       event:
         - push
       ref:
-        - 'refs/heads/DOP-*'
+        - "refs/heads/DOP-*"
 
   - name: publish-staging
     image: plugins/ecr
@@ -47,7 +47,7 @@ steps:
       event:
         - push
       ref:
-        - 'refs/heads/main'
+        - "refs/heads/main"
 
   - name: publish-prod
     image: plugins/ecr
@@ -75,7 +75,7 @@ steps:
       namespace: docs
       release: docs-search-transport
       values: image.tag=git-${DRONE_COMMIT_SHA:0:7},image.repository=795250896452.dkr.ecr.us-east-1.amazonaws.com/docs/${DRONE_REPO_NAME},ingress.enabled=true,ingress.hosts[0]=- docs-search-transport-dev.docs.staging.corp.mongodb.com
-      values_files: ['environments/all.yml', 'environments/dev.yml']
+      values_files: ["environments/all.yml", "environments/dev.yml"]
       api_server: https://api.staging.corp.mongodb.com
       kubernetes_token:
         from_secret: staging_kubernetes_token
@@ -83,7 +83,7 @@ steps:
       event:
         - push
       ref:
-        - 'refs/heads/DOP-*'
+        - "refs/heads/DOP-*"
 
   - name: deploy-staging
     image: quay.io/mongodb/drone-helm:v3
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -8,8 +8,28 @@ trigger:
...
 steps:
+  - name: publish-dev
+    image: plugins/ecr
+    settings:
+      create_repository: true
+      registry: 795250896452.dkr.ecr.us-east-1.amazonaws.com
+      repo: 795250896452.dkr.ecr.us-east-1.amazonaws.com/docs/${DRONE_REPO_NAME}
+      tags:
+        - git-${DRONE_COMMIT_SHA:0:7}
+        - latest
```
rayangler, line 22: (Non-blocking recommendation) I don't think we currently use the `latest` image tag for anything. It might be good to avoid tagging dev images with `latest` in case it causes any unwanted conflicts in the future. This overlap in image tag names was actually an issue on Hapley once upon a time ([PR](https://github.com/mongodb/docs-hapley/pull/97)).

tests/integration/search.test.ts
```diff
@@ -17,32 +17,38 @@ describe('Searching', function () {
...
+        'docs-search-indexes-test',
+        'search-indexes/preprd',
+        client,
+        TEST_DATABASE
+      );
+      const result = await index.load({} as Taxonomy, 'dir:tests/integration/search_test_data/');
+      console.log('index loaded');
+      await index.createRecommendedIndexes();
+      console.log('created recommended indexes');
+      console.log(result);
```
rayangler, line 34: (Just wanted to make sure) Are we purposely leaving these console logs?

.drone.yml
```diff
@@ -8,8 +8,28 @@ trigger:
...
+      access_key:
+        from_secret: ecr_access_key
+      secret_key:
+        from_secret: ecr_secret_key
+    when:
+      event:
+        - push
+      ref:
+        - 'refs/heads/DOP-*'
```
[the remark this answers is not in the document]
seungpark: yep. we can change this to be something like `feature-*` if that seems more dev friendly

tests/integration/search.test.ts
seungpark: was debugging some issues with timeout and createIndex responding without actually creating the index (a 200 response means the request was good, not that the index has finished building).  think it'll be good to leave this for now to debug any flaky tests

.drone.yml
seungpark: good catch! left the `latest` image tag strictly to prod and updated the tags for dev/stage

.drone.yml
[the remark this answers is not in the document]
rayangler: I think this depends on the overall use case for this new environment. If this is intended to only be used for particular branches, I think we should be more particular with the keyword here. If we don't really care if the dev environment can get overwritten whenever, then I'm fine to keep as is!

I'm down to keep as is as long as that's okay

.drone.yml
```diff
@@ -44,6 +64,25 @@ steps:
...
 
+  - name: deploy-dev
+    image: quay.io/mongodb/drone-helm:v3
+    settings:
+      chart: mongodb/web-app
+      chart_version: 4.12.3
+      add_repos: [mongodb=https://10gen.github.io/helm-charts]
+      namespace: docs
+      release: docs-search-transport-dev
+      values: image.tag=git-${DRONE_COMMIT_SHA:0:7},image.repository=795250896452.dkr.ecr.us-east-1.amazonaws.com/docs/${DRONE_REPO_NAME},ingress.enabled=true,ingress.hosts[0]=docs-search-transport-dev.docs.staging.corp.mongodb.com
```
rayangler, line 75: If we're planning on using the image tag from the `publish-dev` step, I think we need to make the image.tag here conform to the same name: `git-${DRONE_COMMIT_SHA:0:7}-dev`. Same with the rename of the staging tag if we plan to add the `-staging` suffix. Otherwise, I don't think the images will be pulled in properly

seungpark: good catch! done

.drone.yml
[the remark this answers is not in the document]
seungpark: i think its fine that the dev env can be overwritten. this is subject to change!

PR #104, opened 2024-04-05
DOP-4323: Boost query matches on taxonomy values

mayaraman19:
### Ticket

DOP-4323

### Notes

Add text matching on facet values. Also fixed a bug in which selected facets were being overwritten by the last facet in the list, as well as added a test to check that the amount of clauses created by `extractFacetFilters()` is the same as the amount of facets passed in through the URL.

Staging link at https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/search/

### README updates

### DIFF
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -181,6 +181,14 @@ export class Query {
       },
     });
 
+    parts.push({
+      text: {
+        query: terms,
+        path: 'facets',
+        score: { boost: { value: 10 } },
+      },
+    });
+
     const compound: Compound = {
       should: constructBuryOperators(parts),
       minimumShouldMatch: 1,
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -184,7 +184,13 @@ export class Query {
     parts.push({
       text: {
         query: terms,
-        path: 'facets',
+        path: [
+          'facets.genre',
+          'facets.programming_language',
+          'facets.target_product',
+          'facets.target_product>atlas>sub_product',
+          'facets.target_product>realm>sub_product',
+        ],
         score: { boost: { value: 10 } },
       },
     });
--- a/src/Query/util.ts
+++ b/src/Query/util.ts
@@ -77,7 +77,6 @@ export const extractFacetFilters = (searchParams: URL['searchParams']): Filter<D
 
     // if subproduct, remove parent
     if (drilldownKeyIdx > -1) {
-      // retrieve parent product from key
       const parent = key.slice(drilldownKeyIdx + 1, key.indexOf('>', drilldownKeyIdx + 1));
       const indexOfParent = queryParamLists[baseFacet].compound.should.findIndex(
         (element: Phrase) => element.phrase.query === parent
--- a/src/Query/util.ts
+++ b/src/Query/util.ts
@@ -2,6 +2,8 @@ import { Filter } from 'mongodb';
 
 import { Document, FacetAggregationStage, FacetOption } from '../SearchIndex/types';
 
+import { Phrase } from './types';
+
 const atomicPhraseMap: Record<string, string> = {
   ops: 'manager',
   cloud: 'manager',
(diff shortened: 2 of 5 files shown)


src/Query/util.ts
```diff
@@ -67,19 +69,28 @@ export const extractFacetFilters = (searchParams: URL['searchParams']): Filter<D
...
       },
     };
+
+    // if subproduct, remove parent
+    if (drilldownKeyIdx > -1) {
+      const parent = key.slice(drilldownKeyIdx + 1, key.indexOf('>', drilldownKeyIdx + 1));
+      const indexOfParent = queryParamLists[baseFacet].compound.should.findIndex(
+        (element: Phrase) => element.phrase.query === parent
+      );
+      if (indexOfParent > -1) queryParamLists[baseFacet].compound.should.splice(indexOfParent, 1);
```
mayaraman19, line 84: This code block assumes that the parent facet is always going to come before the subproduct. While testing nested subfacets in Snooty it seems like this will always be the case (correct me if i'm wrong) - however, this might not always be the case when locally testing `docs-search-transport` since facets are manually being added to the request.

src/Query/index.ts
```diff
@@ -181,6 +181,14 @@ export class Query {
       },
     });
 
+    parts.push({
```
seungpark, line 184: this should be across all different facet properties. ie `facets.target_product`  `facets.target_product>atlas>sub_products`. [these (search for facets)](https://cloud.mongodb.com/v2/5bad1d3d96e82129f16c5df3#/clusters/atlasSearch/Search?collectionName=documents&database=search-staging&indexName=default&view=JsonEdit) are [already being programmatically added onto the search index](https://github.com/mongodb/docs-search-transport/blob/main/src/AtlasAdmin/index.ts#L155), so we should be able to add text operators on these fields

src/Query/util.ts
seungpark: good catch! but i think its safe to say that whatever is passed onto the request should be handled as an OR.

ie. `facets.target_product=atlas&facets.target_product=drivers` should come out to two should clauses.

on the other hand, if the client makes the request of a subfacet `facets.target_product=atlas&facets.target_product=drivers&facets.target_product>atlas_subproduct=atlas-cli`, that is an erroneous request since all sub products of atlas would be included.

TLDR we can remove the base facet logic and include everything in the requested facets.

if the client is including both parent and child facet (when only selecting the child facet), then we should address at client side

seungpark: also a good chance to add unit tests to above!

seungpark (changes requested):

src/Query/index.ts
```diff
@@ -181,6 +181,20 @@ export class Query {
...
+      text: {
+        query: terms,
+        path: [
```
seungpark, line 187: this works for now but i think ideally we would get this programmatically from SearchIndex.trieFacets

tests/unit/Query.test.ts
```diff
@@ -55,4 +56,13 @@ describe('Query', () => {
         ?.value as unknown as number) === 100
     );
   });
+
+  it('should have as many clauses as filters passed into the query', () => {
+    const searchParams = new URLSearchParams(
```
seungpark, line 61: minor nit. can we also add a sub_product facet selection here. also should validate the distinction between compound.shoulds (OR) and separate compounds (AND) - this applies when selecting different base facets

ie. `target_product=atlas & target_product>atlas>sub_product=atlas-cli & programming_language=rust` should be `(target_product = atlas OR target_product>atlas>sub_product = atlas-cli) AND (programming_language = rust)`

(which is true in this case but not really tested. the filters are:)

```
[
  {
    "compound": {
      "should": [
        { "phrase": { "query": "atlas", "path": "facets.target_product" } },
        { "phrase": { "query": "drivers", "path": "facets.target_product" } }
      ]
    }
  },
  {
    "compound": {
      "should": [
        { "phrase": { "query": "go", "path": "facets.programming_language" } }
      ]
    }
  }
]
```

seungpark (approved):
minor comments below but overall it is doing the needful. thanks @mayaraman19 !

src/Query/index.ts
mayaraman19: going to drop in a ticket for this because it's taking longer than expected 😞

sarahemlin (approved):
Hey @mayaraman19!  My test searches all looked good today.  I've also asked Sarah Simpers to test today.  Thank you!

PR #105, opened 2024-04-17
DOP-3934: update custom analyzer to escape $

seungpark:
### Ticket

DOP-3934

### Notes
- Updates custom analyzer to a simple mapping vs regex token filter
- Also removes the [hard coded mapping](https://github.com/mongodb/docs-search-transport/blob/main/src/data/term-result-mappings.ts) from the query as this shouldn't be needed anymore

### README updates

### DIFF
--- a/src/Query/index.ts
+++ b/src/Query/index.ts
@@ -108,18 +108,17 @@ export class Query {
     const parts: Part[] = [];
     const searchPropertyMapping = getPropertyMapping();
 
-    // DOP-3934 - test without having this hardcoded matches
     // if we need to boost for matching slug on an exact rawQuery match
-    // const boostedStrings = strippedMapping[this.rawQuery.trim()];
-    // if (Array.isArray(boostedStrings) && typeof boostedStrings[0] === 'string') {
-    //   parts.push({
-    //     text: {
-    //       path: 'strippedSlug',
-    //       query: boostedStrings,
-    //       score: { boost: { value: 100 } },
-    //     },
-    //   });
-    // }
+    const boostedStrings = strippedMapping[this.rawQuery.trim()];
+    if (Array.isArray(boostedStrings) && typeof boostedStrings[0] === 'string') {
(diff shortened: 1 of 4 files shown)


seungpark:
@sarahemlin this can be tested here https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/search

src/Query/index.ts
```diff
@@ -108,22 +108,30 @@ export class Query {
     const parts: Part[] = [];
     const searchPropertyMapping = getPropertyMapping();
 
+    // DOP-3934 - test without having this hardcoded matches
```
rayangler, line 111: Is the plan to remove the commented code after some time of confirming things work as expected?

src/Query/index.ts
```diff
@@ -108,22 +108,30 @@ export class Query {
...
-    if (Array.isArray(boostedStrings) && typeof boostedStrings[0] === 'string') {
-      parts.push({
-        text: {
-          path: 'strippedSlug',
-          query: boostedStrings,
-          score: { boost: { value: 100 } },
-        },
-      });
-    }
+    // const boostedStrings = strippedMapping[this.rawQuery.trim()];
```
rayangler, line 113: My understanding is that the mapping includes terms that don't have `$` in them, meaning those terms won't be boosted anymore. I didn't see this in the A/C of the original ticket, so I'm just wondering if this is okay to do?

seungpark: good point. i think we should just leave the mapping (this was for testing in qa) and see if there is an overall improvement with one change at a time - changing the mapping, and separately removing this hardcoded mapping

src/Query/index.ts
seungpark: see below response. lets leave this and remove as a separate test

rayangler (approved):

sarahemlin (approved):

PR #107, opened 2024-04-24
DOP-4552: Get facet names programmatically in Query/index.ts

mayaraman19:
### Ticket

DOP-4552 

### Notes

I wanted to leverage the existing `getFacetKeys` function in Search Index, but obtaining the taxonomy and passing it into this function outside `AtlasAdmin` proved a bit cumbersome.

This solution is the main one that comes to mind, but open to any suggestions for better code structure or locally storing the facet key names into a member of `SearchIndex`, similar to [insertTaxonomyIntoSearchIndex](https://github.com/mongodb/docs-search-transport/blob/main/src/AtlasAdmin/index.ts#L155) but to my knowledge the results of this function are only being uploaded online and doesn't exist locally/in a way that's easily accessible to `Query`.

### README updates

### DIFF
--- a/src/SearchIndex/index.ts
+++ b/src/SearchIndex/index.ts
@@ -20,11 +20,11 @@ import {
   RefreshInfo,
   Taxonomy,
   FacetOption,
-  FacetValue,
   FacetAggRes,
   TrieFacet,
-  AmbiguousFacet,
 } from './types';
+
+import { getFacetKeys } from '../AtlasAdmin/utils';
 
 const log = new Logger({
   showTimestamp: true,
@@ -43,6 +43,8 @@ export class SearchIndex {
   unindexable: Collection<DatabaseDocument>;
   trieFacets: TrieFacet;
   responseFacets: FacetOption[];
+  taxonomy: Taxonomy | null;
+  facetKeys: string[];
 
   constructor(manifestSource: string, s3Bucket: string, s3Path: string, client: MongoClient, databaseName: string) {
     this.currentlyIndexing = false;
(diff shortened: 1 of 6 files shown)


src/SearchIndex/index.ts
```diff
@@ -266,3 +267,23 @@ const composeUpserts = (
     };
   });
 };
+
+// similar to atlas admin getFacetKeys but this one is not
+// specific to taxonomy type
+export const getFacetKeysFromResponseFacets = (facets: FacetOption[]): string[] => {
```
seungpark, line 273: can we move this fn to Query/utils and also add a unit test for this

tests/integration/search.test.ts
```diff
@@ -3,6 +3,10 @@ import { Query } from '../../src/Query';
 import { SearchIndex } from '../../src/SearchIndex';
 import { Taxonomy } from '../../src/SearchIndex/types';
 import { MongoClient } from 'mongodb';
+import { sampleFacetOption } from '../resources/utils-data';
+import * as dotenv from 'dotenv';
+
+dotenv.config();
```
seungpark, line 9: thanks for this change! ❤️

src/SearchIndex/index.ts
```diff
@@ -266,3 +267,23 @@ const composeUpserts = (
     };
   });
 };
+
+// similar to atlas admin getFacetKeys but this one is not
```
seungpark, line 271: is it different? CMIIW but getting the same list of keys from the function mentioned in this line

src/SearchIndex/index.ts
seungpark: or not! see comment above

seungpark (commented):
see comment below! seems like we are producing the same result. is it because we don't have the taxonomy at hand at query time? we can store the result of the existing fn `getFacetKeys` in memory and move that fn to be shared by both atlas admin and query build time. this will also prevent this fn having to run with every query - however this will require us to keep an eye out for memory usage to be safe

src/SearchIndex/index.ts
mayaraman19: yes, for comment above, we don't have the taxonomy at hand at query time and acquiring it would either create a large `async`/`await` chain or require passing it through many functions, but I think storing it locally or storing the result of `getFacetKeys` would be useful. I think this is where some stuff about organization I mentioned in Engineering Sync was coming up, but I'll take another crack at this and write everything down.

mayaraman19:
removed the function i added. stored `taxonomy` and result of `getFacetKeys` in `SearchIndex` and passing that into the search functions

seungpark (approved):
this LGTM. i made a quick dashboard to track the memory usage if we need to bump it up in our drone configs: https://grafana.staging.corp.mongodb.com/d/dBFR-ABIz/docs-search-staging-memory?orgId=8&from=now-30d&to=now

PR #109, opened 2024-05-02
DOP-4593: Search endpoint to return HTML format for manifest list

anabellabuckvar:
### Ticket

[DOP-4593](https://jira.mongodb.org/browse/DOP-4593)
Smartling crawler can't follow links in a JSON array, but it can follow HTML links. This endpoint is only being used for Smartling's crawler so this PR converts the format of all manifest links to HTML

### Notes
Sometimes the 'should properly handle incorrect urls in manifests' test passes and sometimes fails on the same code if I rerun jobs, I am not sure why/if it should be investigated more thoroughly.

### README updates

### DIFF
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -346,6 +346,6 @@ export default class Marian {
     };
 
     res.writeHead(200, headers);
-    res.end(response());
+    res.end('hello');
   }
 }
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -347,14 +347,12 @@ export default class Marian {
     }
 
     let manifestList = '';
-    let endList = 'gah';
     const openTags = '<a href=';
-    const hrefClose = '>  ';
-    const closeTags = '  </a>';
-    for (let i = 0; i < this.index.manifests.length; i++) {
-      const manifestUrl = new URL(
-        `${this.index.manifestUrlPrefix}/${this.index.manifests[i].searchProperty}.json`
-      ).toString();
+    const hrefClose = '>';
+    const closeTags = '</a>';
+    const urlPrefix = this.index.manifestUrlPrefix;
+    for (let manifest of this.index.manifests) {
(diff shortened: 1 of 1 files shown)


src/Marian/index.ts
```diff
@@ -333,13 +332,19 @@ export default class Marian {
       return;
     }
 
-    const response = {
-      manifests: this.index.manifests.map((manifest) =>
-        new URL(`${this.index.manifestUrlPrefix}/${manifest.searchProperty}.json`).toString()
-      ),
-    };
+    let manifestList = '';
+    const openTags = '<a href=';
```
seungpark, line 336: minor nitpick. can we wrap these `<a>` tags in either `<div>`s or `<p>` so that they wrap in the page

src/Marian/index.ts
```diff
@@ -333,13 +332,19 @@ export default class Marian {
...
+    const hrefClose = '>';
+    const closeTags = '</a>';
+    const urlPrefix = this.index.manifestUrlPrefix;
+    for (let manifest of this.index.manifests) {
+      const manifestUrl = new URL(`${urlPrefix}/${manifest.searchProperty}.json`).toString();
+      manifestList += openTags + manifestUrl + hrefClose + manifestUrl + closeTags;
```
seungpark, line 342: would also add a `\n` at the end so that the source HTML is also more legible

src/Marian/index.ts
```diff
@@ -333,13 +332,19 @@ export default class Marian {
...
+    }
+
+    const response = '<html><body>' + manifestList + '</body></html>';
```
seungpark, line 345: same new line for after `<html><body>`

seungpark (approved):
minor changes but overall LGTM !

seungpark:
https://docs-search-transport-dev.docs.staging.corp.mongodb.com/manifests looks good!

src/Marian/index.ts
seungpark: this hasn't been addressed but was a minor nitpick !

PR #114, opened 2025-04-17
DOCSP-49373-adding-realm-hard-logic

ltran-mdb2:
### Ticket

https://jira.mongodb.org/browse/DOCSP-49373

### Notes

Adds realm hard logic

### README updates

### DIFF
--- a/src/data/term-result-mappings.ts
+++ b/src/data/term-result-mappings.ts
@@ -228,6 +228,7 @@ const resultMapping: ResultMapping = {
     'reference/operator/aggregation/mergeObjects',
   ],
   'queryable encryption': ['core/queryable-encryption'],
+  'realm': ['atlas/device-sdks/deprecation/'],
 };
 
 // Strips the result mapping of any '/' characters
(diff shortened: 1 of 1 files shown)


sarahemlin (approved):

src/data/term-result-mappings.ts
```diff
@@ -228,6 +228,7 @@ const resultMapping: ResultMapping = {
     'reference/operator/aggregation/mergeObjects',
   ],
   'queryable encryption': ['core/queryable-encryption'],
+  'realm': ['atlas/device-sdks/deprecation/'],
```
seungpark, line 231: lint error!

```suggestion
  realm: ['atlas/device-sdks/deprecation/'],
```

seungpark (changes requested):
minor linting error. otherwise LGTM !

src/data/term-result-mappings.ts
ltran-mdb2: Fixed!

seungpark: ah github issues! there is an extra space here (before realm)

ltran-mdb2: ah I updated it to what I  _think_ is the right spacing - does that look right to you now?

seungpark (approved):

PR #116, opened 2025-04-23
(DOP-5566): Respect order in term-result-mappings.ts

cbush:
### Ticket

[DOP-5566](https://jira.mongodb.org/browse/DOP-5566)

### Notes

This gives more boost to earlier entries in the mapped results so as to respect the order given. If every entry is given the same boost, the ordering in result is not deterministic.

### README updates

### DIFF
--- a/tests/unit/Query.test.ts
+++ b/tests/unit/Query.test.ts
@@ -51,9 +51,11 @@ describe('Query', () => {
     const nonExistingTermQuery = new Query('constructor').getCompound(null, [], sampleFacetKeys);
     ok((nonExistingTermQuery.should[0] as NestedCompound).compound.must[0].text?.score?.boost?.value === undefined);
     const existingTermQuery = new Query('aggregation').getCompound(null, [], sampleFacetKeys);
-    ok((existingTermQuery.should[0] as NestedCompound).compound.must[0].text?.score?.boost !== undefined);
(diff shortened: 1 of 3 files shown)


cbush:
Tested here: usually "update" would be first result, "query" would be second. Now "query" is first as it was specified to be in the mapping.
image

cbush:
Failing test completely unrelated to PR - might need to be updated or removed

tests/unit/Query.test.ts
```diff
@@ -52,10 +52,8 @@ describe('Query', () => {
     ok((nonExistingTermQuery.should[0] as NestedCompound).compound.must[0].text?.score?.boost?.value === undefined);
     const existingTermQuery = new Query('aggregation').getCompound(null, [], sampleFacetKeys);
     ok((existingTermQuery.should[0] as NestedCompound).compound.must[0].text?.score?.boost !== undefined);
-    ok(
-      ((existingTermQuery.should[0] as NestedCompound).compound.must[0].text?.score?.boost
-        ?.value as unknown as number) === 100
-    );
+    ((existingTermQuery.should[0] as NestedCompound).compound.must[0].text?.score?.boost?.value as unknown as number) >=
```
seungpark, line 55: [nit]  can also check for decreasing boost value

seungpark (approved):
LGTM. 

would be extra helpful to outline this in https://github.com/mongodb/docs-search-transport/blob/main/src/data/term-result-mappings.ts

PR #120, opened 2025-05-14
DOP-5316: Only log error if unhandled

rayangler:
### Ticket

DOP-5316

### Notes

**The problem:**

Splunk is logging too many errors and it's unclear why.

**The solution:**

* Added new error messages to `InvalidQuery` throws to make it more specific **why** a query is not being handled.
* Move the error log that Splunk kept picking up around such that it would only log as an error if it's truly unexpected. We should treat `InvalidQuery` errors as "expected" within the context of the application as we only throw them if the user input is bad. We'll still log a warning in case we need to track and debug them.

We'll see how effective this solution is at reducing Splunk error alerts over time after this is deployed to prod.

### README updates

### DIFF
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -120,8 +120,7 @@ export default class Marian {
         res.end(err.message);
         return;
       }
-      log.error(`Unhandled error while handling search from URL ${String(parsedUrl)}:`);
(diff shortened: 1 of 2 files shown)


seungpark (approved):
❤️

PR #122, opened 2025-06-10
DOP-5578: store user query in collection

seungpark:
### Ticket

DOP-5578
This PR aims to store searched terms into a MDB collection in the search db.

### Notes
- Tested this locally and was able to save to `search-test` DB under collection `query`

image

**Note:** unrelated test failing for the Search Index test. Possibly worth rewriting the test suite so we can operate on one set of data

### README updates

### DIFF
--- a/src/Marian/index.ts
+++ b/src/Marian/index.ts
@@ -112,7 +112,7 @@ export default class Marian {
 
     let results: Document[];
     try {
-      results = await this.fetchResults(parsedUrl);
+      results = await this.fetchResults(parsedUrl, req.headers);
     } catch (err) {
       if (err instanceof InvalidQuery) {
         res.writeHead(400, headers);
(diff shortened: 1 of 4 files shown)


src/Marian/index.ts
```diff
@@ -238,7 +238,7 @@ export default class Marian {
     res.end(responseBody);
   }
 
-  private async fetchResults(parsedUrl: URL): Promise<Document[]> {
+  private async fetchResults(parsedUrl: URL, req: http.IncomingHttpHeaders): Promise<Document[]> {
```
rayangler, line 241: Can we rename `req` to `reqHeaders` for consistency with its type? Thought this was the request object itself at first

rayangler (approved):
LGTM with one recommendation

